PR #5, opened 2020-01-12
WIP: github-actions

zimbatm:
This is an attempt to run the self-hosted github-actions runner.

Lessons learned:

* found out most dependencies
* the github registration token refreshes every 30(?) minutes.
* each repo needs their own agent, it would be nice to have org-level
  agents.

Because the agent is designed to unpack in-place we have to copy the
derivation into a runtime directory. It's probably possible to symlink a
subset of the folders instead.

We probably want to use the stateDir instead of the runtimeDir to avoid
having to register the agent on each deployment, due to the expiring
token.

Status
======

The service is deploying and starting.

The current issue is that the github checkout action is breaking with
the following error:

Error:

    [error]Could not load file or assembly 'Runner.Plugins, Culture=neutral, PublicKeyToken=null'. An attempt was made to load a program with an incorrect format.
    [error]Exit code 1 returned from process: file name '/run/github-actions-nixpkgs-fmt/bin/Runner.PluginHost', arguments 'action "GitHub.Runner.Plugins.Repository.v1_0.CheckoutTask, Runner.Plugins"'.

See https://github.com/nix-community/nixpkgs-fmt/runs/385789439


zimbatm:
@ocharles were you able to take this any further?

ocharles:
I'm afraid I haven't had time to try

domenkozar:
I wonder if FHS environment is a more sustainable approach.

zimbatm:
Looking forward to having this merged: https://github.com/NixOS/nixpkgs/pull/116775

Mic92:
I think we can close this since github-actions is in nixpkgs now.

PR #7, opened 2020-01-19
users: add kiwi

Kiwi:
@worldofpeace said I could make this PR and name drop them :smiley: I don't know that I actually need to be in wheel but everyone else was; yay cargo cult. Mostly I want to be able to fix nixpkgs easier... (but I am willing to contribute in any other way I can) so as long as I can do distributed builds(ideal)/local builds I'm happy. My laptop is slow and overheats and crashes quite frequently (I think it might have bad ram...) ... I still try to make lots of PRs to fix stuff though! :P @grahamc trusts me enough that I'm allowed to use the aarch64 server and the ofborg (which I still don't know how to use :| ) has me in the list if that makes me more credible. Thanks :heart:

### DIFF
--- a/users/kiwi.nix
+++ b/users/kiwi.nix
@@ -1,0 +1,25 @@
+{ config, pkgs, lib, ... }:
+
+let
+  userLib = import ./lib.nix { inherit lib; };
+  keys = [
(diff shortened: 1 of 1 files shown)


zimbatm:
Hi Kiwi,

I am sorry, I don't feel comfortable adding you to this machine at the moment. The purpose of this machine is mainly to act as a CI builder (and @ryantm' bot). It currently has some secret tokens embedded that are needed. If it wasn't for those I would gladly give you access to the machine.

I hope it's not too much of a blow and we can find a way to give you access to a proper remote builder.

Kiwi:
How can you not trust a face like this ;)

a picture of Kiwi

It's ok it was a long shot anyway :smile:

zimbatm:
hahaha :D

PR #10, opened 2020-01-28
Add github nixpkgs-swh deployment key

nlewo:
The pipeline `nixpkgs-swh` needs a github token to push to the repository `nix-community/nixpkgs-swh`.
IIUC, the buildkite-agent could provide this token to the pipeline. I would then need to deploy it as a key via the git-crypt keystore.

Todo:
- [x] add the token to the git-crypt store
- [ ] locally build the machine (don't know how to do that with nixops)

@zimbatm could you help me to store my token?


nlewo:
context: I need to push force to the branch `gh-pages` of the `nixpkgs-swh` repository.

Actually, it doesn't seems easy to push with a GitHub token. I think I should use a SSH keypair instead.

build01/buildkite.nix
```diff
@@ -13,5 +13,14 @@
       pkgs.nix
       pkgs.gzip
     ];
+
+    hooks.environment = ''
+      #!/usr/bin/env bash
+      set -euo pipefail
+
+      if [[ "$BUILDKITE_PIPELINE_SLUG" == "nixpkgs-swh" ]]; then
```
zimbatm: you might want to limit access to the master branch. not sure if that's enough to prevent third-party users from stealing the token though.
```suggestion
      if [[ "$BUILDKITE_PIPELINE_SLUG" == "nixpkgs-swh" && "$BUILDKITE_BRANCH" == master ]]; then
```

nlewo:
Thanks to a @zimbatm protip, a token can do the job actually (no https password is needed since GitHub supports the token as username for `https` remotes).

zimbatm:
can you give me your GPG public key so I can add you to git-crypt?

nlewo:
`wget https://keybase.io/lewon/key.asc` (key fingerprint `ACD8 5D1A 5A1C 90D5 7FDE  59D0 533B AFFC 5320 B368`)

zimbatm:
I got a different fingerprint (B31F 0457 ...). Can you send me the key through another channel?

nlewo:
@zimbatm I finally switched to a github SSH deployment key which is more scoped than the `lewo-ci` github token.
Also, I just added the secret in `/var/keys` with the read permission sets for the `buildkite-agent` user. Since the same user is used to run buildkite hooks and to run pipelines, I don't think it's relevant to set a hook to copy the secret.

zimbatm:
sounds good to me!

PR #15, opened 2020-04-05
ci: add basic nix and cachix support

zimbatm:
pre-build the machines so they can be pulled from the cache

### DIFF
--- a/deploy
+++ b/deploy
@@ -2,10 +2,17 @@
 #! nix-shell ./shell.nix -i bash
 set -euo pipefail
 
+options=(
+  --option extra-substituters "https://nix-community.cachix.org"
+  --option binary-cache-public-keys "cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= nix-community.cachix.org-1:mB9FSh9qf2dCimDSUo8Zy7bkq5CX+/rkCWyvRCYg3Fs="
+)
+
 mkdir -p state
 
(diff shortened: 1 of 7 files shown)


ryantm:
I think we should modify the deploy script to use the subsituter and trusted-public-keys of nix-community.

ryantm:
Also, does this mean we need to give our git-crypt secrets to GitHub?

Mic92:
It should be actually possible to separate the secrets from the build process and not having the secrets in the nix store.

zimbatm:
At the moment the `CACHIX_SIGNING_KEY` is added to the repo secrets and only available to GitHub Actions. To populate the cache you have to push a branch on the repo or get it merged into master.

deploy
```diff
@@ -2,10 +2,17 @@
 #! nix-shell ./shell.nix -i bash
 set -euo pipefail
 
+options=(
+  --option extra-substituters "https://nix-community.cachix.org"
+  --option binary-cache-public-keys "cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= nix-community.cachix.org-1:mB9FSh9qf2dCimDSUo8Zy7bkq5CX+/rkCWyvRCYg3Fs="
```
zimbatm, line 7: I don't know why but nixops seems to ignore those options

ryantm: Is your user trusted?

zimbatm: Yeah. Do you mind if the PR gets merged regardless? I don't think it hurts in its current state.

ryantm: I don't mind.

zimbatm: thanks! :)

PR #18, opened 2020-04-17
Update nixpkgs to current master

jonringer:
for the pypi releases, this prevents creating python-3.6 drv-names, which is causing a lot of the packages to be filtered out.


zimbatm:
thanks!

zimbatm:
@jonringer I had to revert as it breaks the NixOS config.

zimbatm:
the buildkite agent and ACME certs have changed

jonringer:
Yea, I'm going to create new PR, to just have it point to the $XDG_CACHE_HOME from @ryantm  suggestion

PR #19, opened 2020-04-17
Use XDG_CACHE_HOME to point to nixpkgs

jonringer:
supercede: https://github.com/nix-community/infra/pull/18

just need the `nix-env -qa ` command to get the latest nixpkgs, I don't need to update the nixpkgs checkout for the entire build :)


jonringer:
cc @ryantm @zimbatm

zimbatm:
we are actually updating everything. see #nix-community on IRC

zimbatm:
there is also an issue with the secrets currently that prevents non-trusted users to test the deployment

jonringer:
ah, I will wait then. Prehaps this PR will not be needed then

zimbatm:
yeah all good!

PR #21, opened 2020-04-17
hydra: add Simple NixOS MailServer project

nlewo:


zowoq:
https://hydra.nix-community.org/project/simple-nixos-mailserver

@nlewo Currently the check interval for these jobsets is 30 and 60 seconds, could you increase it to 5 minutes please? It would be helpful for debugging and maintenance on the infra if it wasn't constantly evaluating.

nlewo:
The 30 seconds interval is for MR. If we increase this value, NixOS mailserver contributors would have to wait more time to get the CI status of their MRs. So, I think it would be nice if we could preserve this interval.

For debugging purpose, do you think it would be possible to temporary disable the NixOS mailserver project.

zowoq:
> The 30 seconds interval is for MR. If we increase this value, NixOS mailserver contributors would have to wait for more time to get the CI status of their MRs.

Yes, this is why I asked for it to be increased to 5 minutes and not 30 minutes like the other hydra projects.

> For debugging purpose, do you think it would be possible to temporary disable the NixOS mailserver project.

No. I actually need it for debugging and even if I do disable it I still need to remember to re-enable it.

nlewo:
Done by https://gitlab.com/simple-nixos-mailserver/nixos-mailserver/-/merge_requests/364

zowoq:
Thank you.

PR #22, opened 2020-05-01
secrets: only remove suffix "\n" for the buildkite token

nlewo:
The SSH private key file needs to have a last empty new
line. Otherwise, SSH fails with:

    Load key "/path/to/key": invalid format

### DIFF
--- a/deployment.nix
+++ b/deployment.nix
@@ -1,5 +1,21 @@
+with builtins;
+
 let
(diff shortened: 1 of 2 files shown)


nlewo:
Hm, I just hope this will not break some existing secrets :/

zimbatm:
We used to not remove the \n at the end so it will be fine

Issue #24, opened 2020-06-10
Host marvin-mk2


timokau:
I'd like to [move](https://github.com/NixOS/rfcs/pull/30#issuecomment-640126206) the [marvin-mk2](https://discourse.nixos.org/t/rfc-marvin-mk2-the-pull-request-bot/7547) bot ([repo](https://github.com/timokau/marvin-mk2)) onto nix-community infrastructure before starting to test it on nixpkgs.

It should consume very little resources. Would that be possible? If so, how should I go about it?

adisbladis:
Fixed! We're nov hosting it :)

timokau:
> Also, all the current people who have access to the secrets will need to be okay with you having access.

Should I open an issue about this? Having to nag one of you every time I'd like to deploy an update is not ideal, especially since I can't control when a deploy happens :)

ryantm:
Yes, it's probably best to open another issue, then I can ping all the people with access to get their okay.

PR #25, opened 2020-06-10
build01: add initial marvin-mk2 configuration

timokau:
Adding marvin-mk2, see https://github.com/nix-community/infra/issues/24.

This is a work in progress.


build01/marvin-mk2.nix
```diff
@@ -0,0 +1,65 @@
...
+  systemd.services.marvin-mk2 = {
+    description = "marvin-mk2 service";
+    enable = true;
+    path = marvin-mk2-runtimeenv;
+    environment.XDG_CONFIG_HOME = "/var/lib/marvin-mk2";
+    environment.XDG_CACHE_HOME = "/var/cache/marvin-mk2";
+    # FIXME
+    environment.GH_KEY_FILE = "";
+    # FIXME
+    environment.GH_APP_ID_FILE = "";
```
Mic92, line 44: By the way you can use EnvironmentFile to pass secrets from a file.

build01/marvin-mk2.nix
```diff
@@ -0,0 +1,65 @@
...
+  users.users.marvin-mk2.packages = [
+  ];
+  users.groups.marvin-mk2 = { };
+  users.users.marvin-mk2 = {
+    useDefaultShell = true;
+    uid = userLib.mkUid "mmk2";
+    extraGroups = [ "marvin-mk2" ];
+  };
+  nix.trustedUsers = [
+    "marvin-mk2"
```
Mic92, line 32: I am curious why it needs to be a trusted user.

build01/marvin-mk2.nix
zimbatm: That's quite nice. Potentially it means that the secret could even be owned by root isn't it?
I had a quick look at the [EnvironmentFile documentation](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=) but it doesn't mention what user it will use to read the file.

build01/marvin-mk2.nix
```diff
@@ -0,0 +1,65 @@
...
+  ];
+
+  systemd.services.marvin-mk2 = {
+    description = "marvin-mk2 service";
+    enable = true;
+    path = marvin-mk2-runtimeenv;
+    environment.XDG_CONFIG_HOME = "/var/lib/marvin-mk2";
+    environment.XDG_CACHE_HOME = "/var/cache/marvin-mk2";
```
zimbatm, line 40: small nitpick, you can use:
```suggestion
    environment.XDG_CACHE_HOME = "%C";
```

See https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Specifiers

build01/marvin-mk2.nix
timokau: Oh, nice! Thanks.

timokau: Having it owned by root might be nice, but since its exposed in the environment anyways the security benefits are probably marginal.

build01/marvin-mk2.nix
timokau: It absolutely doesn't, just a leftover from copy & pasting r-ryantm's definition. In my defense, I did say its work in progress ;) Thanks for catching!

build01/marvin-mk2.nix
timokau: I'm not sure why/if its even necessary to set these in the first place. What do they default to? Currently marvin-mk2 doesn't need any cache/config anyways, but that may change.

(again, the settings are included since I based this of r-ryantm)

timokau:
Thanks @adisbladis for all your help :tada:

PR #27, opened 2020-06-16
add CNAME for *.nix-community.org to github pages

zimbatm:
What do you think of this?

It would allow to, for example, have https://nur.nix-community.org/ instead of https://nix-community.github.io/nur-search/

/cc @Mic92

### DIFF
--- a/terraform/cloudflare_nix-community_org.tf
+++ b/terraform/cloudflare_nix-community_org.tf
@@ -52,3 +52,19 @@ resource "cloudflare_record" "mumble-AAAA" {
   value   = "2a03:3b40:fe:ab::1"
   type    = "AAAA"
 }
+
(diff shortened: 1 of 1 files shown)


zimbatm:
One potential issue is that it becomes a vector of attack against https://hydra.nix-community.org/

Mic92:
Maybe it could be set explicit per project?

nlewo:
Yep, I also think this would be better to set it by project.
It would actually be nice to maintain a list of supported projects. This list could be used to generate domain names, but could also be used to generate a static webpage describing all supported projects.

zimbatm:
Ok, the first iteration is to have a list of subdomains that can then be taken by repos by changing their CNAME value.

PR #26, opened 2020-06-13
marvin-mk2: update

timokau:
The version in #25 was mostly to get the deployment working. I think this version is ready for a test run on nixpkgs. It currently only reacts to PRs where it is summoned in the initial PR message.

- slash commands instead of @mentions
- some fixes
- training wheels removed (no toy commands anymore, does not run on issues)
- logs exceptions instead of crashing the entire server
- some fixes, improved messages

PR #28, opened 2020-06-16
Open port 3001 for marvin

timokau:
https://github.com/nix-community/infra/pull/26#issuecomment-644935561


ryantm:
I already did this in a somewhat crappier way in https://github.com/nix-community/infra/commit/131c0d13b032135ba6835cbf997afa2afbcf27d3

timokau:
Should be just as good, there is little reason to change the port in the future.

PR #29, opened 2020-06-16
marvin: update

timokau:
What would a first deploy be without a typo. Only change is
https://github.com/timokau/marvin-mk2/pull/14 ("Fix status
terminology").

We should get the instructions in order before asking people to test it. @ryantm, this time it really should be a simple deploy :)


ryantm:
Deployed!

timokau:
Thank you!

PR #36, opened 2020-10-06
remove nixos wiki

Mic92:
This is no longer maintained. People should go to nixos.wiki instead.


zimbatm:
thanks, I already pushed and deployed the change

PR #37, opened 2020-10-21
hydra: increase Postgresql cache size

nlewo:
Hydra was not responding anymore...


nlewo:
note this change is currently deployed.

ryantm:
@nlewo I think we better merge it then so people don't overwrite it by accident.

PR #40, opened 2020-11-25
Fix CI

ryantm:
This is a continuation of the work in https://github.com/nix-community/infra/commit/bb579a7db165d4205669b122303338414b22371c

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -7,8 +7,11 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - uses: cachix/install-nix-action@v12
-    - uses: numtide/cachix-action@only-cache
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
+    - uses: cachix/cachix-action@v8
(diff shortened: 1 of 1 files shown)


Mic92:
I can also recommend dependabot for updating actions: https://github.com/Mic92/nix-update/blob/master/.github/dependabot.yml

zimbatm:
it would be good to be pure, but that will fix the immediate problem

ryantm:
By "pure" do you mean not rely on the ever-changing nixos-unstable?

zimbatm:
yeah, especially since we're tracking nixos-unstable, it would be good to pin it in the code

ryantm:
I think this was the problem:

https://github.com/nix-community/infra/blob/5eee3864f3361baf3ea4a770514494b667a342c3/build01/hardware-configuration.nix#L7

zimbatm:
indeed, always use modulesPath

zimbatm:
https://github.com/nix-community/infra/pull/42

PR #42, opened 2020-11-25
ci: fix purity

zimbatm:


.github/workflows/nix.yml
```diff
@@ -7,11 +7,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - uses: cachix/install-nix-action@v12
-      with:
-        nix_path: nixpkgs=channel:nixos-unstable
     - uses: cachix/cachix-action@v8
       with:
         name: nix-community
-        extraPullNames: nixpkgs-update
-        skipPush: true
```
ryantm, line 16: Is `skipPush` implied by not having the secrets provided?

.github/workflows/nix.yml
```diff
@@ -7,11 +7,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - uses: cachix/install-nix-action@v12
-      with:
-        nix_path: nixpkgs=channel:nixos-unstable
     - uses: cachix/cachix-action@v8
       with:
         name: nix-community
-        extraPullNames: nixpkgs-update
```
ryantm, line 15: I was hoping to add this to avoid extra building work, but I'm fine if you don't want it.

zimbatm: good point, restored

.github/workflows/nix.yml
zimbatm: good question. I pushed an update that sets the signingKey instead.

ryantm: This was originally using your numtide branch with the name "only-cache". I think we wanted to skip pushing to avoid pushing secrets into the cache.

zimbatm: Too late :sweat:

zimbatm: I will flush the cache

zimbatm: Actually, the CI is not decoding the secrets!

PR #50, opened 2021-01-18
build01: add telegraf profile

Mic92:


profiles/telegraf.nix
```diff
@@ -0,0 +1,84 @@
...
+          disk.tagdrop = {
+            fstype = [ "tmpfs" "ramfs" "devtmpfs" "devfs" "iso9660" "overlay" "aufs" "squashfs" ];
+            device = [ "rpc_pipefs" "lxcfs" "nsfs" "borgfs" ];
+          };
+        };
+        outputs = {
+          prometheus_client = {
+            listen = ":9273";
+            metric_version = 2;
+            # TODO http auth
```
Mic92, line 63: I could start by scraping this with my existing prometheus setup - I have IRC alerts set up there. We can also deploy a Prometheus service later to both servers.

profiles/telegraf.nix
```diff
@@ -0,0 +1,84 @@
+{ pkgs, lib, config, ... }: let
+  isVM = lib.any (mod: mod == "xen-blkfront" || mod == "virtio_console") config.boot.initrd.kernelModules;
+  secrets = import ../secrets.nix;
+in {
+  config = {
+    systemd.services.telegraf.path = [ pkgs.nvme-cli ];
+
+    # TODO http auth
+    #deployment.keys."telegraf-secret-env".text = secrets."telegraf-secret-env";
```
Mic92, line 9: You seem to add deployment keys to the same file. Is this the preferred way?

profiles/telegraf.nix
```diff
@@ -0,0 +1,84 @@
...
+        options = [ "NOPASSWD" ];
+      }];
+    }];
+    # avoid logging sudo use
+    security.sudo.configFile = ''
+      Defaults:telegraf !syslog,!pam_session
+    '';
+    # create dummy file to avoid telegraf errors
+    systemd.tmpfiles.rules = [
+      "f /var/log/telegraf/dummy 0444 root root - -"
```
Mic92, line 81: This is used for healtchecks. Timers can write updates to this file and prometheus will check if timers run successfully or at all.

profiles/telegraf.nix
zimbatm: Is it necessary to hide the service behind auth? I don't think it's exporting anything sensitive.

zimbatm (approved):
sounds good on principle. welcome!

profiles/telegraf.nix
ryantm: We have all the deployment.keys in one place, so we can avoid importing secrets.nix in multiple places.

profiles/telegraf.nix
Mic92: Not really.

profiles/telegraf.nix
Mic92: dropped from the profile.

profiles/telegraf/default.nix
```diff
@@ -0,0 +1,83 @@
...
+                    while ("${pkgs.zfs}/bin/zpool status" | getline) {
+                        if ($1 ~ /pool:/) { printf "zpool_status,name=%s ", $2 }
+                        if ($1 ~ /state:/) { printf " state=\"%s\",", $2 }
+                        if ($1 ~ /errors:/) {
+                            if (index($2, "No")) printf "errors=0i\n"; else printf "errors=%di\n", $2
+                        }
+                    }
+                }
+              '')
+              (pkgs.runCommandNoCC "mdraid-health" {
```
Mic92, line 45: Added this one as well. Not fully tested yet since I don't run software raids.

PR #52, opened 2021-01-27
build build02 in ci

Mic92:


ryantm (approved):

default.nix
```diff
@@ -11,4 +11,6 @@ in
 pkgs.nix-community-infra // rec {
   build01 = importNixOS ./build01/configuration.nix "x86_64-linux";
   build01-system = build01.system;
+  build02 = importNixOS ./build02/configuration.nix "x86_64-linux";
+  build02-system = build02.system;
```
zimbatm, line 15: Recently I found out that you can now also do `pkgs.nixos [ ./build02/configuration.nix ]`. Much shorter to type :-p

zimbatm: importNixOS returns the eval output, which contains the options, config, ... Since there is no recurse into attrs, we also expose the system here so it gets built.

`pkgs.nixos` outputs the top-level derivation so that would not be needed anymore.

Mic92: I tried here but it seems to not evaluate the nixos system: https://github.com/nix-community/infra/pull/53/files

PR #53, opened 2021-01-28
simplify ci configuration

Mic92:


default.nix
```diff
@@ -9,8 +9,6 @@ let
     };
 in
 pkgs.nix-community-infra // rec {
-  build01 = importNixOS ./build01/configuration.nix "x86_64-linux";
-  build01-system = build01.system;
-  build02 = importNixOS ./build02/configuration.nix "x86_64-linux";
-  build02-system = build02.system;
+  build01 = pkgs.nixos [ ./build01/configuration.nix ];
```
Mic92, line 12: I don't think this actually build the system: https://github.com/nix-community/infra/runs/1783510216

zimbatm: hmm, this might require another `pkgs.recurseIntoAttrs`

PR #54, opened 2021-02-25
add zfs profile

Mic92:


adisbladis (approved):

profiles/common.nix
```diff
@@ -4,7 +4,8 @@
 
   imports = [
     ./security.nix
-    ./telegraf
+    ../services/telegraf
+    ./zfs.nix
```
ryantm, line 8: Build02 is not using zfs. It is using ext4.

Mic92: Yes. It's just a noop in this case.

PR #56, opened 2021-03-04
move users configuration to seperate profile

Mic92:
Untested, lets see if it evaluates.


Mic92:
I want to prepare an common kexec image for booting so that we don't reinvent the wheel for every new installation and also make recovery easier.

Mic92:
No rebuilds == success

PR #58, opened 2021-03-04
add sshd service

Mic92:


services/sshd.nix
```diff
@@ -0,0 +1,12 @@
+{ config, lib, pkgs, ... }:
+{
+  services.openssh = {
+    enable = true;
+    passwordAuthentication = false;
```
Mic92, line 5: I believe everyone is using ssh keys?

adisbladis: Yes, in fact we've always had `services.openssh.passwordAuthentication = false;` set (in `profiles/security.nix`).

Mic92: Ok. I would still keep it here as well because I am importing this file into the kexec image.

adisbladis: Sounds good to me! Just wanted to point out that we are already banning password logins :)

zimbatm (approved):

PR #59, opened 2021-03-06
add build03

Mic92:


build03/configuration.nix
```diff
@@ -0,0 +1,39 @@
...
+  boot.loader.grub.devices = [ "/dev/nvme0n1" "/dev/nvme1n1" ];
+  boot.loader.grub.enable = true;
+  boot.loader.grub.version = 2;
+  boot.loader.grub.extraConfig = ''
+    # for mdraid 1.1
+    insmod mdraid1x
+  '';
+
+  networking.nix-community = {
+    ipv4.address = "135.181.218.169";
```
Mic92, line 27: @ryantm I would like to unify network configuration for all hosts. Any concerns switching build02 to this and away from dhcpd?

ryantm: My concern is the non predictable name might not be eth0. Is there a way to check that without doing it?

Mic92: @ryantm right. You would need to unset this and do `./deploy --boot --include build03 --force-reboot`

PR #60, opened 2021-03-06
declarative hydra users

Mic92:
Code is untested.


build01/hydra.nix
```diff
@@ -135,14 +138,22 @@ in
       path = with pkgs; [ hydra-unstable netcat ];
       script = ''
         set -e
-        export HYDRA_ADMIN_PASSWORD=$(cat ${cfg.services.hydra.adminPasswordFile})
+        while IFS=; read -r user role password email; do
+          if [[ "$user" == "admin" ]]; then
+            export HYDRA_ADMIN_PASSWORD="$password"
+          fi
+          opts=("$user" "--role" "$role" "--password" "$password")
```
grahamc, line 145: Not a lot better but hydra-create-user has:

```
  [--password-hash SHA1-HASH]
```

Mic92: Could we export those hashes from the current database?

Mic92: done.

PR #61, opened 2021-03-07
move things around a bit

zimbatm:
use ./services for service instances and ./roles for shared module
configurations.

This will make it easier to move services between hosts.


Mic92 (changes requested):
Don't forget kexec.nix in profiles

ryantm:
Since we are renaming refactoring, I guess it's time for a little bikeshedding.

At work, we use the words `modules` for NixOS modules that potentially could be upstreamed, and `roles` for shared configs. Roles is shorter than profiles and what is used by Chef for the same thing.

Mic92 (approved):

ryantm (approved):

PR #63, opened 2021-03-08
roles/security: allow sudo without password

zimbatm:
Encourage users to SSH as their own users. User accounts were not able
to `sudo` as they don't have a password associated to them.

### DIFF
--- a/roles/security.nix
+++ b/roles/security.nix
@@ -1,6 +1,8 @@
 { config, pkgs, lib, ... }:
 
 {
+  # Make sure that the firewall is enabled, even if it's the default.
+  networking.firewall.enable = true;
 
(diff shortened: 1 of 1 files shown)


adisbladis:
This is already possible with SSH agent forwarding.

Mic92:
> This is already possible with SSH agent forwarding.

zimbatm does not trust the other admins to not use his ssh agent..

ryantm (approved):

zimbatm:
> zimbatm does not trust the other admins to not use his ssh agent.

I trust you :) Assuming that your ssh-agent contains all of your keys, a hacker that gains root access on the boxes now has lateral access to other resources like all our private github repos.

PR #70, opened 2021-03-24
improve recovery documentation switch to dhcp/

Mic92:


roles/hetzner-network.nix
```diff
@@ -37,14 +24,13 @@ in {
     networking.dhcpcd.enable = false;
     systemd.network = {
       enable = true;
-      networks."eth0".extraConfig = ''
+      networks."ethernet".extraConfig = ''
         [Match]
-        Name = eth0
+        Type = ether
         [Network]
+        DHCP = ipv4
```
Mic92, line 31: Let's see if this fixes this issue. At least it is less code, which is also good.

zimbatm (approved):

PR #75, opened 2021-03-29
Migrate from `queued-build-hook` to `cachix watch-store`

adisbladis:
We've had instability with the former, and it turns out Cachix has
gained some functionality that's better for our use case.

### DIFF
--- a/roles/nix-community-cache.nix
+++ b/roles/nix-community-cache.nix
@@ -1,51 +1,17 @@
-{ config, pkgs, ... }:
+{ config, pkgs, lib, ... }:
 let
-  postBuildHook = pkgs.writeScript "post-build-hook.sh" ''
-    #!${pkgs.runtimeShell}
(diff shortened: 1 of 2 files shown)


roles/nix-community-cache.nix
```diff
@@ -1,51 +1,17 @@
...
-
-  systemd.services.queued-build-hook = {
-    description = "Post-build-hook service";
+  systemd.services.cachix-watch-store = {
+    description = "Cachix store watcher service";
     wantedBy = [ "multi-user.target" ];
-    after = [ "network.target" "queued-build-hook.socket" ];
-    requires = [ "queued-build-hook.socket" ];
+    after = [ "network.target" ];
+    path = [ pkgs.nix ];
```
zimbatm, line 11: Let's use the same version of nix everywhere:
```suggestion
    path = [ config.nix.package ];
```

PR #84, opened 2021-04-15
add rust-for-linux hydra job

Kloenk:
This adds hydra ci jobs for the rust-for-linux linux tree.

I spoke with @Mic92 about adding this to the nix-community hydra.

### DIFF
--- a/services/hydra/declarative-projects.nix
+++ b/services/hydra/declarative-projects.nix
@@ -30,5 +30,12 @@
       description = "Packages for Redox";
       homepage = "https://github.com/nix-community/redoxpkgs";
     };
+    rust-for-linux = {
+      displayName = "Rust For Linux";
+      inputValue = "https://github.com/rust-for-linux/nix";
+      specFile = ".hydra/spec.json";
+      description = "Linux Kernel with Rust support";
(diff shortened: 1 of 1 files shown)


Mic92:
I like it and it is good marketing for the power of nix :)

services/hydra/declarative-projects.nix
```diff
@@ -30,5 +30,12 @@
       description = "Packages for Redox";
       homepage = "https://github.com/nix-community/redoxpkgs";
     };
+    rust-for-linux = {
+      displayName = "Rust For Linux";
+      inputValue = "https://github.com/rust-for-linux/nix";
+      specFile = ".hydra/spec.json";
```
nlewo, line 36: Note you will need to rename it in your repository (where it is `specs.json`).

nlewo (approved):
LGTM as well!

services/hydra/declarative-projects.nix
Kloenk: Oh, thanks :-)

Kloenk: Fixed in my repo, thank's for that.

Issue #85, opened 2021-04-17
Move marvin-2k to build03


Mic92:
I want to move marvin-2k to build03. I saw that it currently uses webhooks as callbacks.
What domain did you configure as a callback, if it is hardcoded can we make a new subdomain just for marvin-2k?

timokau:
The bot was using a builder-specific domain. It is now configured to use the new subdomain (#95). Sorry for the late response. I saw the notification and meant to respond soon, but somehow it ended up being 17 days later.

I would rather not be the bottleneck for things like this. I would like marvin-mk2 to be "from the community, for the community". Contribution isn't easy right now, but as a start I could transfer ownership of the GitHub app to the nix-community organization. What do you think? (CC @zimbatm)

zimbatm:
Good idea @timokau

PR #95, opened 2021-05-02
marvin-2k: add marvin-2k.nix-community.org

zimbatm:
Add a layer of indirection to make the migration to build03 easier.

Please update the GitHub Webhook to point to this address first.

/cc @timokau @Mic92

part of #85

### DIFF
--- a/build01/configuration.nix
+++ b/build01/configuration.nix
@@ -15,6 +15,7 @@
 
     ../roles/common.nix
     ../roles/hetzner-network.nix
+    ../roles/nginx.nix
 
     ../services/marvin-mk2.nix
   ];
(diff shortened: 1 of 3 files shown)


zimbatm:
https://marvin-2k.nix-community.org is returning a 404 but it seems to be returning the same on port :3001

timokau:
I have configured the bot to use the new URL for the webhook. Not that it matters, but is there a reason you picked marvin-2k instead of marvin-mk2?

It might be a good idea to transfer ownership of the GitHub app to the nix-community organization. See #85.

Mic92:
> I have configured the bot to use the new URL for the webhook. Not that it matters, but is there a reason you picked marvin-2k instead of marvin-mk2?

false memory? we can also rename it.

timokau:
I don't mind either way, we would just have to synchronize it.

PR #97, opened 2021-05-08
update nixpkgs

Mic92:


roles/common.nix
```diff
@@ -14,8 +14,6 @@
   environment.systemPackages = [
     # for quick activity overview
     pkgs.htop
-    # for users with TERM=xterm-termite
-    pkgs.termite.terminfo
```
Mic92, line 18: cc @zimbatm I had to remove it for now since termite is broken in nixpkgs-unstable.

PR #110, opened 2021-08-17
add build04

Mic92:
this is an aarch64 server.


deployment.nix
```diff
@@ -163,4 +163,13 @@ in
         permissions = "0600";
       };
     };
+
+  build04 =
```
Mic92, line 167: This will break our CI, which is x86_64.

zimbatm:
nice!

Mic92:
There is an interesting bug in nix-build-uncached. It just seems to ignore aarch64. However this is currently to our advantage as we can just add the server without breaking anything.

PR #124, opened 2021-09-25
migrate to sops-nix

Mic92:
This gets rid of slow nixops key deployment and git-crypt in long-term

Would you like to use sops-nix with pgp or age?
cc @nlewo @adisbladis @flokli @ryantm @zimbatm @timokau


.sops.yaml
```diff
@@ -0,0 +1,37 @@
+keys:
+  - &mic92 age17n64ahe3wesh8l8lj0zylf4nljdmqn28hvqns2g7hgm9mdkhlsvsjuvkxz
```
Mic92, line 2: This can be either a pgp fingerprint or an age key.

Mic92: For gpg: https://github.com/Mic92/sops-nix#2a-generate-a-gpg-key-for-yourself
For age: https://github.com/Mic92/sops-nix#2a-generate-a-ssh-and-age-key-for-yourself

nlewo:
I never used age. @Mic92 Is it possible to store age key in a yubikey?

Mic92:
> I never used age. @Mic92 Is it possible to store age key in a yubikey?

If you want to use your yubikey than gnupg is still the better choice. I heard newer yubikeys also work with ed25519 (the crypto algorithms used by age) https://github.com/str4d/age-plugin-yubikey 
But I don't think this has been integrated with sops.

.sops.yaml
ryantm: Oh, is there not support for using SSH keys directly?

Mic92: No the upstream Age maintainer did not want this when reviewing the sops PR that added age support. However we build an easy to use tool that does the conversion.

Mic92:
Here is an example what adding a gpg key looks like: https://github.com/Mic92/dotfiles/blob/f94773404d6ded8ac4e338f8cbaa2d2b5b914e57/nixos/.sops.yaml#L2

zimbatm:
Can we stick to GPG for now since we all already have a key for it? That should allow making the migration transparent.

Mic92:
> Can we stick to GPG for now since we all already have a key for it? That should allow making the migration transparent.
> 
> Age looks cool but I need to look into how to backup/restore the private key properly.

Sure everyone can use the key format they prefer.

Mic92:
Please share your pgp or age key fingerprint here. I tried to use the git-crypt one, but gpg says those are invalid for most users.

nlewo:
@Mic92 my pgp key fingerprint is `B31F04575C4561F58DE02C9EFC6C2E50B00E9861`.

Mic92:
> @Mic92 my pgp key fingerprint is `B31F04575C4561F58DE02C9EFC6C2E50B00E9861`.

Please upload your public key to https://github.com/nix-community/infra/tree/master/keys
I could not find your key on either ubuntu or sks key server.

PR #129, opened 2021-10-03
switch from nixops to morph

Mic92:
nixops is broken and requires a pinned channel.
For our use case morph seems simpler as it is stateless


ryantm (changes requested):
I'm pretty sure this is missing necessary secrets changes.

ryantm (approved):
Oh sorry, I didn't notice you already merged the sopsnix branch.

PR #136, opened 2021-11-01
hydra: upload cachix directly

Mic92:
This no longer requires stopping hydra when doing nix-gc to avoid
race conditions in cachix-watch store.

### DIFF
--- a/services/hydra/default.nix
+++ b/services/hydra/default.nix
@@ -92,13 +92,10 @@ in
       ];
     };
 
-    #services.nginx.virtualHosts = {
-    #  "hydra.nix-community.org" = {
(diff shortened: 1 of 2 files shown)


services/hydra/default.nix
```diff
@@ -111,6 +108,14 @@ in
       usersFile = hydraUsersFile;
       extraConfig = ''
         max_output_size = ${builtins.toString (8 * 1024 * 1024 * 1024)}
+
+        <runcommand>
+        command = ${pkgs.writeShellScript "cachix-upload" ''
```
Mic92, line 113: I am tested this on my own server for a while

nlewo (approved):

PR #140, opened 2021-11-12
build04: add hercules-ci.nix

Mic92:


ci.nix
```diff
@@ -13,6 +13,8 @@ pkgs.nix-community-infra // rec {
   build01-system = build01.system;
   build02 = importNixOS ./build02/configuration.nix "x86_64-linux";
   build02-system = build02.system;
-  build03 = importNixOS ./build02/configuration.nix "x86_64-linux";
+  build03 = importNixOS ./build03/configuration.nix "x86_64-linux";
   build03-system = build03.system;
+  build04 = importNixOS ./build04/configuration.nix "aarch64-linux";
```
roberth, line 18: Now that there's a second node, you need to configure the binary cache in binary-caches.json, so that drvs can be shared between nodes.

Mic92: Yeah. We noticed this the hard way.

roberth: I'd like to make it more flexible, but right now the focus is on more high level aspects like a new nix file (including flakes support) and more "typical" CI/CD features.

Mic92: It's got green https://hercules-ci.com/github/nix-community/infra/jobs/13 eventually.

ci.nix
```diff
@@ -13,6 +13,8 @@ pkgs.nix-community-infra // rec {
...
+  build04-system = build04.system;
 }
```
roberth, line 20: Actually it'd be ok for this file to be `default.nix`. 

> At the top-level of your repository, the agent will search for a Nix file to evaluate in the following order: nix/ci.nix, ci.nix or default.nix

 -- https://docs.hercules-ci.com/hercules-ci-agent/evaluation/
 
 It also documents the traversal of attributes.
 
 Next release series, 0.9.x, is updating this to include flake.nix support among other things.

roberth: That said, renaming this to `ci.nix` is of course fine if CI is the whole purpose of this file.

Mic92: Yes, making the intent of this file clear, was my main motivation.

PR #176, opened 2022-01-16
build01: add winter as trusted user

winterqt:


roles/builder/users.nix
```diff
@@ -22,6 +22,11 @@ let
       trusted = true;
       keys = ./keys/jtojnar;
     };
+
+    winter = {
+      trusted = true;
```
zimbatm, line 27: @Mic92 trusted user basically gives root to the machine. Are we comfortable with that?

winterqt: I followed in the footsteps of the aarch64 builder. If you'd like me to negate that, let me know.

Mic92: I had a chat with @andir. Let's go one step back. @winterqt maybe you can give us feedback on how not beeing a trusted user affects your workflow. It might be not possible to use our server as a remote builder in this case?

winterqt: I do not believe that I need to be a trusted user to make use of this server, as most of my needs are building Nixpkgs related things such as NixOS tests. I will let you know if that changes though, if that would help.

winterqt: @Mic92 Nevermind:
```
> error: you are not privileged to build input-addressed derivations
```

This was when building a NixOS configuration (for tests). I can use the aarch64 builder for this, so it's not an issue, but figured I'd let you know.

Mic92: Oh. This is a bummer. Thanks for letting us know.

winterqt: Nope.

```
$ nix --store ./s build -f thelounge.nix --argstr system x86_64-linux -L --show-trace
error: creating directory '/s': Read-only file system
error: unexpected EOF reading a line

       … while reading the response from the build hook
```

winterqt:
Is it possible that my status as a non-trusted user on this machine could be reevaluated? Currently I can't use this builder at all (to build packages or run NixOS tests) because I'm not trusted.

I'm already a trusted user on the aarch64 builder, plus as of recently I am a Nixpkgs committer.

Thanks!

zimbatm:
Is it impossible to fix the NixOS tests to work as untrusted users? I found this explanation, read it three times, and still don't get it:
https://github.com/nixos/nix/blob/6776e65fd960e25b55d11a03324f9007b6dc2a0b/src/libstore/daemon.cc#L565-L597

winterqt:
I believe that comment only applies when using this as a remote builder -- doing all my work remotely (cloning Nixpkgs, etc.) would probably let me run tests, but that's also not preferable, and by that point I'll continue just using the aarch64 box (and I really don't mind doing so).

See https://github.com/NixOS/nix/issues/2789 for a better explanation as to why remote builds require a trusted user, at least for now.

PR #185, opened 2022-02-08
add user

zowoq:
Having access to a faster builder would be helpful for stuff like golang updates.

I would need access to remote builds and nixos tests for this to be useful so I've set `trusted = true`.

_____

Also, could an admin please send me an invite to the `nix-community` org? I'm planning to contribute to a few of the repos and not needing to bother with the github actions approval dance would be nice.


Mic92:
Your key has been deployed.

Mic92:
invite sent.

zowoq:
Thanks @Mic92!

PR #210, opened 2022-04-16
Remove marvin-mk2

timokau:
I have [discontinued](https://github.com/timokau/marvin-mk2/issues/34#issuecomment-1100656280) the bot. I'm not very familiar with flakes and the current setup of this repository, so it would be good if someone could review my changes in some detail. I modified the `flake.lock` by hand for example, which you are probably not supposed to do.

Feel free to remove my privileges on this repository as well. Less people with access can only be good for security.


Mic92:
Thanks! Applied in https://github.com/nix-community/infra/pull/212

PR #213, opened 2022-04-18
Enable aarch64 emulation on Hydra systems

adisbladis:
I would like to build Emacs for aarch64 and have it in the binary cache.

Note that this is only intended for the Emacs distribution itself and not it's thousands of packages.

### DIFF
--- a/services/hydra/default.nix
+++ b/services/hydra/default.nix
@@ -140,12 +140,14 @@ in
       buildMachines = [
         {
           hostName = "localhost";
-          systems = [ "x86_64-linux" "builtin" ];
+          systems = [ "x86_64-linux" "aarch64-linux" "builtin" ];
           maxJobs = 8;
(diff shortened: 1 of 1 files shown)


services/hydra/default.nix
```diff
@@ -140,13 +140,15 @@ in
...
           hostName = "localhost";
-          systems = [ "x86_64-linux" "builtin" ];
+          systems = [ "x86_64-linux" "aarch64-linux" "builtin" ];
           maxJobs = 8;
           supportedFeatures = [ "nixos-test" "big-parallel" "kvm" ];
         }
       ];
     };
 
+    boot.binfmt.emulatedSystems = [ "aarch64-linux" ];
```
Mic92, line 150: We have build04 as a remote builder with native aarch64. Maybe just use that. It's an oracle arm machine.

PR #217, opened 2022-05-04
Add review-bot for build01

ckiee:
https://github.com/nix-community/review-bot

Discussion of this was on Matrix a while ago. TL;DR is you ping it with PR numbers and it runs `nixpkgs-review` on them. 

The `review-bot` flake includes an untested nixosModule so this should be easy. 

##### TODO
- [ ] Fill in the `roles/review-bot.nix` file (@Mic92)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ result-*
 # Terraform
 .terraform
 .terraform.*
+
+# direnv
+.direnv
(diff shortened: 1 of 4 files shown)


ryantm:
Cool. I had the thought to break out the report that r-ryantm makes on it's PRs into its own code path. If that happened, we could call that instead of just nixpkgs-review.

zimbatm:
@ckiee are you still interested in pushing this forward? Let me know if I can unblock you

PR #224, opened 2022-05-15
get rid of git-crypt

Mic92:


zimbatm:
good :)

zowoq:
Are the keys in `.git-crypt` still used or can they be removed as well?

zimbatm:
They can be removed

PR #233, opened 2022-07-07
nixpkgs-update: use fetcher -> fifo -> worker model

ryantm:


Mic92:
anything else needed here?

ryantm:
I had to rollback this change because it was force pushing to PRs in a bad way. Been sick a lot so no time to debug.

PR #249, opened 2022-08-23
builder/users: add fgaz

fgaz:
I'd like to request access to the builder. I maintain a fair amount of packages but my current machine isn't very powerful.

I set `trusted = true` because I'd like to run nixos tests too:
```
$ cd nixpkgs
$ git grep fgaz nixos/tests/ | wc -l
9
```

I'm already a trusted user in the aarch64 machine
https://github.com/nix-community/aarch64-build-box/blob/a1a761e93469949a33a308a0aee064f970ec2cc7/users.nix#L157-L160


fgaz:
Thanks!

PR #261, opened 2022-09-07
hercules-ci-agent: Use master

roberth:
Update hercules-ci-agent to `master` to help fix nix-community/dream2nix-nodejs-auto, https://github.com/hercules-ci/support/issues/67.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -59,9 +59,66 @@
         "type": "github"
       }
     },
+    "flake-parts_2": {
+      "inputs": {
+        "nixpkgs": [
+          "hercules-ci-agent",
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1657102481,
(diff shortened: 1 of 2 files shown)


flake.lock
```diff
@@ -366,7 +484,7 @@
     },
     "sops-nix": {
       "inputs": {
-        "nixpkgs": "nixpkgs_5",
+        "nixpkgs": "nixpkgs_6",
```
zimbatm, line 487: do you mind garbage-collecting those instances of nixpkgs a bit?

roberth: I've added one nixpkgs and Nix renumbered them.

I guess `follows` might be an option, if that's what you mean? But I don't think you'll be quite happy with that either, because I can't really promise hercules-ci-agent to be quite as portable between nixpkgs versions as you may hope, especially because the `nix` package does not have C++ api stability.

zimbatm: Ok, forget it then

DavHau:
Thanks a lot for this

PR #285, opened 2022-10-04
ci: remove effects deployment

zowoq:
Probably not much need for effects deployment with hourly auto upgrades. (Though maybe we should set up something to monitor the auto upgrades?)

Looks like the ssh key is in sops but someone else will need to remove that.


zimbatm (approved):
less code == good

zimbatm:
actually, how do we get notified if a host fails to deploy?

zowoq:
I don't think there is any monitoring for `autoUpgrade` at the moment.

PR #304, opened 2022-10-28
add terraform deploy action

zowoq:
cc @Mic92 @zimbatm 

Opened this for comment but TBH it's probably not worth merging here as mergify is more useful as terraform changes don't happen very often. However I think this workflow would work well for managing the nixos.org hydra.

The atlantis package that I mentioned on matrix isn't actually suitable for use in public repos, there are a few alternatives but they are either fairly new, not compatible with our terraform packaging, targeting enterprise ($$$), etc.

### DIFF
--- a/.github/workflows/terraform.yml
+++ b/.github/workflows/terraform.yml
@@ -1,0 +1,45 @@
+name: "terraform"
+
+# for security should only run on push to bors branches (staging/trying)
+on:
+  push:
+    branches:
+      - staging
+      - trying
+
+jobs:
+  terraform-deploy:
+    if: github.repository == 'nix-community/infra'
+    runs-on: ubuntu-latest
+    env:
+      SOPS_AGE_KEY: ${{ secrets.SOPS_AGE_KEY }}
+      TF_IN_AUTOMATION: true
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dorny/paths-filter@v2
+        id: changes
+        with:
+          filters: |
+            terraform:
+              - 'terraform/**'
+      - uses: cachix/install-nix-action@v18
+        if: steps.changes.outputs.terraform == 'true'
+        with:
+          extra_nix_config: |
+            experimental-features = nix-command flakes
+            access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}
+      - name: init
+        if: steps.changes.outputs.terraform == 'true'
+        run: nix develop -c ./terraform/deploy terraform init -input=false
+      - name: validate
+        if: steps.changes.outputs.terraform == 'true'
+        run: nix develop -c ./terraform/deploy terraform validate
+      - name: fmt
+        if: steps.changes.outputs.terraform == 'true'
+        run: nix develop -c ./terraform/deploy terraform fmt -check
+      - name: plan
+        if: steps.changes.outputs.terraform == 'true' && github.ref == 'refs/heads/trying'
+        run: nix develop -c ./terraform/deploy terraform plan -input=false
+      - name: apply
+        if: steps.changes.outputs.terraform == 'true' && github.ref == 'refs/heads/staging'
+        run: nix develop -c ./terraform/deploy terraform apply -auto-approve -input=false
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,5 @@
+"merge with bors":
+  - terraform/**/*
+
+"terraform":
+  - terraform/**/*
--- a/.github/workflows/labels.yml
+++ b/.github/workflows/labels.yml
@@ -1,0 +1,14 @@
+name: "labels"
+
+on:
+  pull_request_target:
+
+jobs:
+  labels:
+    runs-on: ubuntu-latest
+    if: github.repository == 'nix-community/infra'
+    steps:
+      - uses: actions/labeler@v4
+        with:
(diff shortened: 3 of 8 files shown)


.github/workflows/terraform.yml
```diff
@@ -0,0 +1,38 @@
...
+      # TODO: fmt files
+      #- name: fmt
+      #  run: nix develop -c ./terraform/deploy terraform fmt -check
+      - name: plan
+        if: github.ref == 'refs/heads/trying'
+        run: nix develop -c ./terraform/deploy terraform plan -input=false
+      - name: apply
+        # master == remote state
+        if: github.ref == 'refs/heads/staging'
+        run: nix develop -c ./terraform/deploy terraform apply -auto-approve -input=false
```
zimbatm, line 38: For more fancy approaches, see https://github.com/hashicorp/setup-terraform#usage

It would be nice to display the planned changes on the PR before merging.

.mergify.yml
```diff
@@ -1,10 +0,0 @@
-pull_request_rules:
-  - name: automatic merge on CI success
-    conditions:
-      - check-success=ci/hercules/evaluation
-      - check-success=ci/hercules/derivations
-      - author=github-actions[bot]
-    actions:
-      merge:
-        method: merge
-      delete_head_branch: {}
```
zimbatm, line 10: I didn't get why that is mutually exclusive with the terraform plan?

.github/workflows/terraform.yml
```diff
@@ -0,0 +1,38 @@
+name: "terraform"
+
+# for security should only run on push to bors branches (staging/trying)
+on:
+  push:
+    branches:
+      - staging
+      - trying
```
zimbatm, line 8:
```suggestion
      - trying
    paths:
      - terraform/**/*
```
to test but it might help reduce the amount of work the CI is doing. See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push

zimbatm (commented):
why not. Atlantis is one of those projects that looks amazing, but never end up using. GHA can get you 90% there and doesn't require deploying yet another service.

.mergify.yml
zowoq: If `terraform-deploy` is a github required status check for merging and it only runs on staging/trying branches mergify wouldn't be able to merge the PR as the status wouldn't be green. Same for restricting it to `terraform/**` path, non-terraform PRs also wouldn't be green.

```diff
diff --git a/.github/workflows/null.yml b/.github/workflows/null.yml
new file mode 100644
index 0000000..928d347
--- /dev/null
+++ b/.github/workflows/null.yml
@@ -0,0 +1,20 @@
+name: "null"
+
+on:
+  push:
+    branches:
+      - staging
+      - trying
+    paths-ignore:
+      - "terraform/**"
+  pull_request:
+    paths-ignore:
+      - "terraform/**"
+
+jobs:
+  terraform-deploy:
+    if: github.repository == 'nix-community/infra'
+    runs-on: ubuntu-latest
+    steps:
+      - name: skip
+        run: echo "no changed terraform files, skipping"
diff --git a/.github/workflows/terraform.yml b/.github/workflows/terraform.yml
index 3314741..268eded 100644
--- a/.github/workflows/terraform.yml
+++ b/.github/workflows/terraform.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - staging
       - trying
+    paths:
+      - "terraform/**"
 
 jobs:
   terraform-deploy:

```

Could work around that problem with something like this that always sets a green status except for terraform PRs. However it wouldn't work for update flake PRs, as a PR from a github action can't trigger another action we'd need the update PRs to be submitted by a bot account.

.github/workflows/terraform.yml
zowoq: Oops, I had typed a response for this but mustn't have clicked the button.

`bors` will always post a comment with a link to the CI run, two comments on every run may be a bit confusing or noisy. Also posting comments on PRs is slightly complicated by the bors workflow as it's a push event, not a pull request so we'd also need to tell it which PR it originated from.

Mic92:
Looking forward to this PR. I will steal it for my university nixos repo.

.mergify.yml
zowoq: I guess for this repo it isn't really necessary that we make it a required check, only downside is that something could get merged by accident without being applied. Labelling the relevant PRs automatically would probably prevent that.

zowoq:
Ah, looks like I didn't test it properly. Seems paths-ignore doesn't do what I thought it did, have to find another way.

.github/workflows/terraform.yml
```diff
@@ -0,0 +1,48 @@
...
+  terraform-deploy:
+    if: github.repository == 'nix-community/infra'
+    runs-on: ubuntu-latest
+    env:
+      # TODO: add to repo secrets
+      SOPS_AGE_KEY: ${{ secrets.SOPS_AGE_KEY }}
+      TF_IN_AUTOMATION: true
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dorny/paths-filter@v2
```
zowoq, line 20: I would have preferred not to depend on another action but handling the diff/filter logic ourselves would be a maintenance burden. We could run it unconditionally on every `bors merge` but then any terraform problems would block non-terraform PRs.

zowoq:
Forgot to undraft the PR when I added the secrets.

PR #329, opened 2022-11-28
build02: remove hercules

zowoq:
Might save a bit of memory, don't really need it on both build02 and build03. 

Also aligns the hercules and hydra config (build03/build04).

### DIFF
--- a/build02/configuration.nix
+++ b/build02/configuration.nix
@@ -5,7 +5,6 @@
     ../roles/common.nix
     ../roles/hardware/hetzner-amd.nix
     ../roles/hetzner-network.nix
-    ../roles/hercules-ci
     ../roles/nginx.nix
     ../roles/raid.nix
     ../roles/remote-builder/aarch64-build04.nix


Mic92:
I think those memory issues mostly come from parallel evaluations inside nixpkgs-update. Maybe this build02 could be added as a remote builder to build03?

zowoq:
> I think those memory issues mostly come from parallel evaluations inside nixpkgs-update.

I'm still a bit suspicious of https://github.com/NixOS/nixpkgs/commit/c2b898da7623a39e3c9b6265d311fe182aa526f0#diff-7c7ca80928136cfc73a02d5b28350bd900e331d6d304857053ffc9f7beaad576, a few others have mentioned that they're getting oom on the PR. Maybe we could try adjusting core/max-jobs on build02 (should help with parallel evaluations as well) but not really sure what would work best with the updater?

> Maybe build02 could be added as a remote builder to build03?

We could try this but I don't know if hercules supports it or if the agent is required?

> dream2nix project ...

Yeah, this looks like it's the most common and longest job, I'll keep checking it to see if it get worse.

PR #330, opened 2022-11-29
various cleanups

zowoq:

### DIFF
--- a/roles/common.nix
+++ b/roles/common.nix
@@ -30,9 +30,6 @@
   # Just disable it since we are using telegraf to monitor raid health.
   systemd.services.mdmonitor.enable = false;
 
-  # enable "sar" system activity collection
-  services.sysstat.enable = true;
-
   # Make debugging failed units easier
   systemd.extraConfig = ''
     DefaultStandardOutput=journal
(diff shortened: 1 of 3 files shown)


roles/common.nix
```diff
@@ -30,9 +30,6 @@
   # Just disable it since we are using telegraf to monitor raid health.
   systemd.services.mdmonitor.enable = false;
 
-  # enable "sar" system activity collection
-  services.sysstat.enable = true;
```
zowoq, line 34: Not sure if anyone is looking at the output of this, doesn't look like telegraf collects it?

roles/zfs.nix
```diff
@@ -5,8 +5,6 @@
     autoSnapshot.monthly = 1;
     autoScrub.enable = true;
   };
-  # zfs is not available for latest lts kenrel
-  boot.zfs.enableUnstable = true;
```
zowoq, line 9: I think this is fine to remove?

roles/common.nix
Mic92: I think so.

roles/zfs.nix
Mic92: yeah.

PR #339, opened 2022-12-12
shell: use toString for sopsPGPKeyDirs

zowoq:

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -5,7 +5,7 @@
 with pkgs;
 mkShellNoCC {
   sopsPGPKeyDirs = [
-    "./keys"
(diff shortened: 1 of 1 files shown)


shell.nix
```diff
@@ -5,7 +5,7 @@
 with pkgs;
 mkShellNoCC {
   sopsPGPKeyDirs = [
-    "./keys"
+    "${toString ./.}/keys"
```
zimbatm, line 8: Does `sopsPGPKeyDirs` not take paths directly?

Mic92: I think it would, but I am too lazy to test :)

PR #343, opened 2022-12-19
Pull in common configuration from srvos

Mic92:
I am trying to get rid of nixos configuration that I copy-n-paste from project to project.
This way there is only one place where I need to fix stuff.


README.md
```diff
@@ -121,29 +121,41 @@ $ inv deploy --hosts build02 reboot --hosts build02
 ```
 
 ## Install/Fix system from Hetzner recovery mode
-1. Install kexec image from Hetzner recovery system as described in [kexec.nix](roles/kexec.nix) and boot into it
 
-2. Format and/or mount all filesystems to /mnt:
+1. Copy your ssh key to the recovery system so that the kexec image can re-use it.
+
+``` console
+yourmachine> ssh-copy-id build0X.nix-community.org
```
Mic92, line 128:
```suggestion
yourmachine> ssh-copy-id root@build0X.nix-community.org
```

PR #350, opened 2022-12-31
flake: remove build04

zowoq:
Fix eval error: https://hercules-ci.com/github/nix-community/infra/jobs/667


Mic92:
Only issue is that I cannot use `build04.nix-community.org` as a flake reference when building from commandline.

zowoq:
Oh, I should have checked that, sorry.

Could we drop the domain and just use the hostname in the flake?

Mic92:
@zowoq I think this should work https://github.com/nix-community/infra/commit/d4b728135746e2fbb70379b59b9efa7a673fd3b3
at least this is what nixos-rebuild does.

Issue #321, opened 2022-11-24
Rules for administrators


zimbatm:
@adisbladis
@flokli
@grahamc
@Mic92
@nlewo
@ryantm
@zimbatm
@zowoq

I propose to write down a few rules for the administrators of this project for transparency reasons and to make it more democratic.

### Responsibilities as an administrator

Our mission is to support the users of the org, and the Nix project in general.

This is a voluntary effort, on a best-effort basis. Things that are good to do are:
* keep the systems updated and patched
* reply to user requests
* host new services that could help the project

### Quorum of 5

To reduce the attack surface on the project, I propose limiting the number of administrators to 5 people.

In the current configuration, I propose to keep Mic92 and zowoq because they are actively managing the infra, and ryantm so he can deploy his bot. That leaves 2 other people from the existing list.

### How to become an administrator

In order to become an administrator, we ask that you already started contributing to this repo through PRs, and that you are trustworthy. Trust is built on top of personal relationships and past behaviour.

Right now we trust everybody on this list, so that makes it easy for you to come back.

---
Did I miss anything? Does that sound reasonable?

zowoq:
Need to clarify what "administrator" means, admin access to this repo, sudo on the boxes and access to the sops secrets?

`nix-community` owners have commit access to this repo anyway ...

Who's responsible for dealing with the hetzner and oracle accounts?

zimbatm:
Good question.

I think that the easiest would be to give administrators all the accesses:
* sudo on the boxes
* access to the sops secrets
* access to SaaS/Cloud accounts that we use (using sops shared secrets if we can't issue individual accounts)
* ownership on the org so they can fix teams and repo ownerships, set repo and org-level secrets, ...

Right now, the servers are funded by @ryantm and @numtide, so some access might be limited there. In the long run, I would like to create an umbrella organization to hold those and the domain.

Mic92:
The oracle account was provided by me for aarch64.

zowoq:
> To reduce the attack surface on the project, I propose limiting the number of administrators to 5 people.

> I think that the easiest would be to give administrators all the accesses ... sudo on the boxes .. ownership on the org

Will org ownership be limited to just the five administrators? Owners can give themselves sudo and read some secrets straight off the boxes, etc.

Don't really see a way around this without switching to manual deployments or moving the infra repo out of this org.

zimbatm:
Yes, the ownership should only be given to the admins. We're assuming that everybody is acting in good faith.

[...]

PR #351, opened 2022-12-31
add treefmt-nix to repository

Mic92:


Mic92:
bors merge

zowoq:
I've disable the treefmt check, it isn't enforceable unless we move to a PR only workflow.

TBH I think this PR have have been discussed first before it was merged.

zimbatm:
What kind of discussion do you want to have around code formatting?

zowoq:
- Which formatters do we want enabled.
- Should it be a CI check.
- Should it be blocking.
- Should it only check the files that changed in the commit.
- Does our workflow need to change.

zimbatm:
I think a lot of this is based on usage. Can we not try it for a bit and see how it goes?

* Which formatters do we want enabled: one per language. Pick the most commonly used.
* Should it be a CI check: yes, it's the role of CI to test code.
* Should it be blocking: probably, but it can be annoying.
* Should it only check the files that changed in the commit: no because GitHub doesn't run CI on every commit.
* Does our workflow need to change: run `nix flake check` before pushing code?

Mic92:
Pull requests are now mandatory for making changes to this repository.

zowoq:
> Pull requests are now mandatory for making changes to this repository.

Does this apply to admins/owners or just to people with write permission? 

> Which formatters do we want enabled: one per language. Pick the most commonly used.

We currently have three enabled for nix. deadnix is very tedious so I want it gone: https://github.com/nix-community/infra/pull/369 

> Should it be blocking: probably, but it can be annoying.

I'd really prefer that we keep formatting and build/eval separate so we can ignore it if we want to and it doesn't break everything.

> Should it only check the files that changed in the commit: no because GitHub doesn't run CI on every commit.

I should have said PR instead of commit.

zowoq:
> > Pull requests are now mandatory for making changes to this repository.
> 
> Does this apply to admins/owners or just to people with write permission?

Seems this just applies to write permission as I can still push to master: https://github.com/nix-community/infra/commit/d3775368953c0a8741840342cb0d3ae564b90bd8

zowoq:
1) Check gets enabled.
2) Check gets broken on master so I disable it and say it should have been discussed first.
3) I list topics for discussion when asked.
  - (At this point I assumed we'd have a discussion and people would get a chance to approve of the changes before they happened.)
4) Check get re-enabled.
5) "Pull requests are now mandatory" workflow change.

We went from no formatting and a permissive workflow to very strict formatting and a restricted workflow. 

I think it's fair to say that these aren't insignificant changes and that they warranted discussion and consensus, that this didn't happen is bit surprising and disappointing. Frankly it makes me concerned about similar situations occurring in the future.

Also "Pull requests are now mandatory" isn't actually true so CI can still end up in a broken state.

Mic92:
My reasoning behind adding the check was that I sometimes end up accidentally pushing to master when I actually wanted to make a PR instead (because I was on the wrong branch). I think with bors it becomes less of a burden to make all changes through a pull request. I am sorry that the rule I applied was not effective.

zowoq:
> My reasoning behind adding the check was that I sometimes end up accidentally pushing to master when I actually wanted to make a PR instead (because I was on the wrong branch). I think with bors it becomes less of a burden to make all changes through a pull request. I am sorry that the rule I applied was not effective.

Sorry but this doesn't address my concerns.

In the interest of not wasting peoples time I'll be blunt, apologies if it comes across as rude, that isn't my intention.

- Why was this PR merged without consensus from the maintainers? I would have expected that it be left open for several days to give everyone a chance to comment.

- After I listed topics for discussion why was the check re-enabled so quickly? I would have expected a discussion and being given the chance to approve of it being re-enabled.

- Why was "pull requests are now mandatory" announced without consensus from the maintainers?

I'm not saying that any of this was done in bad faith but honestly doesn't feel like working collaboratively.

zimbatm:
@zowoq, I think the question is what attitude we want towards the project.

In my view, the person that does the work gets precedence in the general case. That's why I was happy to see you introduce many changes to the project, and didn't require strong consensus. It's going in the right general direction and I think the work you're doing is great.

If you want you can call this change controversial and require a majority vote of the infra team, per #321. Or make a counter proposal, maybe there is a middle ground that would work for you better? Or we can try it like this for a week and see how it goes? The infra will keep changing over time so it's not like this is permanent.

zowoq:
> I think the question is what attitude we want towards the project.

What I keep thinking is that the way all of this was done was unnecessary.

There wasn't anything gained by rushing this through.

With only a few people to consult who all seem to be responsive: why not give them a chance to do so?

> In my view, the person that does the work gets precedence in the general case.

I wouldn't consider these changes to be "general case", I think that they are very different from anything that's been merged in the last few months.

> If you want you can call this change controversial and require a majority vote of the infra team ... Or make a counter proposal, maybe there is a middle ground that would work for you better?

Should have be able to do this before it was merged and if I had said I wanted to make a counter proposal I should have been given some time to do so.

> Or we can try it like this for a week and see how it goes?

We were doing this, it broke and I disabled it.

It has been re-enabled but the underlying problem (pushing to master) hasn't been resolved.

I would have said it was still problematic and objected to it being re-enabled if I had been given a chance to do so.

I think I've found right settings to resolve it and maybe even make bors mandatory if we want (no manual merges).

It would have been much better to have discussed and been able to implement this if necessary as part of the original PR or before it was re-enabled.

Instead I've spent my time trying to figure out how to make it work even when I don't really agree with it because it seems I have no choice but to either fix it or accept that people can push unformatted changes to master and break CI.

> The infra will keep changing over time so it's not like this is permanent.

Yes, however I don't get the feeling that arguing against the something that has already been merged is likely to lead to a satisfying outcome.

I'm still concerned that we'll have similar occurrences in the future.

PR #352, opened 2023-01-01
treefmt: remove shell

zowoq:
`ci.sh` seems a bit redundant. (also needs aarch64 and x86_64 to build)

Removing shell from treefmt as now it isn't needed for anything in the repo.


zowoq:
Only other scripts in the repo are two `deploy`, guess we could suffix them with `.sh`?

PR #361, opened 2023-01-05
misc cleanups

zowoq:

### DIFF
--- a/build01/configuration.nix
+++ b/build01/configuration.nix
@@ -25,9 +25,6 @@
   networking.hostName = "build01";
   networking.hostId = "d2905767";
 
-  # Emulate armv7 until we have proper builders
-  boot.binfmt.emulatedSystems = [ "armv7l-linux" ];
-
   systemd.network.networks."10-uplink".networkConfig.Address = "2a01:4f8:13b:2ceb::1/64";
 
   system.stateVersion = "20.03";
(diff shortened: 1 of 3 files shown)


roles/users.nix
```diff
@@ -18,18 +18,4 @@ in
 
   # No mutable users
   users.mutableUsers = false;
-
-  # Assign keys from all users in wheel group
-  # This is only done because nixops cant be deployed from any other account
-  users.extraUsers.root.openssh.authorizedKeys.keys = lib.unique (
```
Mic92, line 24: So effectively you are disabling the root login via ssh? We are still depending on it here: https://github.com/nix-community/infra/blob/6cea644066371ce3bcd107ad3a49bd4bc2b3679a/tasks.py#L21

zowoq: Dropped this commit.

build01/configuration.nix
```diff
@@ -25,9 +25,6 @@
   networking.hostName = "build01";
   networking.hostId = "d2905767";
 
-  # Emulate armv7 until we have proper builders
-  boot.binfmt.emulatedSystems = [ "armv7l-linux" ];
```
Mic92, line 29: I think I want to bring riscv64-linux in though... arm 32-bit is probably a dead-end for NixOS anyway.

PR #366, opened 2023-01-06
merge nix-community.org in this repo

zimbatm:
This repo is our central hub with the website and config.

It will make it easier to do docs change like in #321


zowoq:
https://github.com/nix-community/infra/settings/pages
The custom domain dns check is still in progress, should it be taking this long? 

____

I've tried tinkering with this locally and seeing what jekyll serve but I can't seem to find the right combination of settings to get rid of "infra".

infra

PR #368, opened 2023-01-08
introduce the rulebook

### LINKED ISSUE #321: Rules for administrators

zimbatm:
This is the start of a constitution for the project.

Closes #321

### DIFF
--- a/docs/rulebook.md
+++ b/docs/rulebook.md
@@ -18,8 +18,9 @@ There are currently two classes of members in the project:
 
 ### Team
 
-Anybody who contributes code to one of our project is considered a community
-member.
+Anybody that fills these criteria automatically becomes a members:
+* Has at least 10 contributions to projects that are in the org
+* The last contribution is younger than 2 years old
 
 ### Responsibilities
 
--- a/docs/rulebook.md
+++ b/docs/rulebook.md
@@ -97,4 +97,4 @@ participation in the project.
 
 > FIXME: changing the rulebook should only be possible from a community vote.
 
-Changing this rulebook requires a quorum majoriy from the administrators.
+Changing this rulebook requires a quorum majority from the administrators.
--- a/docs/rulebook.md
+++ b/docs/rulebook.md
@@ -1,0 +1,100 @@
+# Rulebook / constitution
+
+This is a small document that defines how this community is working.
+
+## Project role
+
+Provide a shared development environment and infrastructure for the Nix and
+NixOS ecosystem, strictly for non-profit purposes.
+
+## Members
+
+There are currently two classes of members in the project:
+
+* Community members - participants of our community.
+* Administrators - members who own and administer the project.
+
+## Community members
+
+### Team
+
+Anybody who contributes code to one of our project is considered a community
+member.
+
+### Responsibilities
+
+Community members are asked to help keep a good spirit focused on building and
+shipping software. Everybody is human and shall be treated with compassion and
+respect.
+
+### Power
+
+> FIXME: this is not available yet.
+
+Community members might create a referendum. Any referendum that has more than
+10 votes gets submitted to the community at large for voting. The voting
+result will be enacted on a majority vote.
+
+## Administrators
+
+### Team
+
+* @Mic92
+* @adisbladis
+* @ryantm
+* @zimbatm
+* @zowoq
+
+### Role
+
+The role of the administrators is to support the members of the org, and the
+Nix project in general.
+
+### Responsibilities
+
+On top of the community members responsibilities, the administrators are also
+responsible for serving the community to the best of their ability and
+availability.
+
+They provide services such as:
+* keep the systems updated and patched
+* reply to user requests
+* host new services that could help the project
+* on-board new projects
+* resolve conflict
+
+### Quorum of 5 rule
+
+To reduce the attack surface on the project, the administrative team size is
+limited to 5 people.
+
+### Power
+
+The administrators have the power to change any aspect of the project.
+
+Controversial changes require a majority vote. Any administrator can call a
+change controversial :-)
+
+> FIXME: we don't have a voting mechanism available yet.
+
+The administrators might delegate some of the decision making process as a
+vote to the wider community.
+
+The administrators are the only "owners" of the github organization.
+
+### Become an administrator
+
+> FIXME: Our intention is to make that process more democratic in the future.
+
+New administrators are selected by the current team.
+
+Administrators would typically be selected from their previous activity and
+participation in the project.
+
+## Rules
+
+### Changing the rules
+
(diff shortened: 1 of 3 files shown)


Mic92 (approved):

docs/rulebook.md
```diff
@@ -0,0 +1,100 @@
...
+Community members might create a referendum. Any referendum that has more than
+10 votes gets submitted to the community at large for voting. The voting
+result will be enacted on a majority vote.
+
+## Administrators
+
+### Team
+
+* @Mic92
+* @adisbladis
```
Mic92, line 43: Is @adisbladis still actively doing stuff here? There is also winter, who would is maintaning the darwin build box: https://github.com/winterqt/darwin-build-box/commits/main

zimbatm: Not a lot, but member rotation can happen separately. This PR is about introducing the mechanisms.

docs/rulebook.md
```diff
@@ -0,0 +1,100 @@
+# Rulebook / constitution
+
+This is a small document that defines how this community is working.
+
+## Project role
+
+Provide a shared development environment and infrastructure for the Nix and
```
zowoq, line 7: Please change this doc to soft wrap, it's a bit annoying to try to keep hard wrap consistent and github markdown ignores it anyway.

docs/rulebook.md
```diff
@@ -0,0 +1,100 @@
...
+shipping software. Everybody is human and shall be treated with compassion and
+respect.
+
+### Power
+
+> FIXME: this is not available yet.
+
+Community members might create a referendum. Any referendum that has more than
+10 votes gets submitted to the community at large for voting. The voting
+result will be enacted on a majority vote.
```
zowoq, line 36: I think this section needs a lot more detail before it can be included here.

```suggestion

```

docs/rulebook.md
```diff
@@ -0,0 +1,100 @@
...
+Administrators would typically be selected from their previous activity and
+participation in the project.
+
+## Rules
+
+### Changing the rules
+
+> FIXME: changing the rulebook should only be possible from a community vote.
+
+Changing this rulebook requires a quorum majoriy from the administrators.
```
zowoq, line 100:
```suggestion
Changing this rulebook requires a quorum majority from the administrators.
```

docs/rulebook.md
```diff
@@ -0,0 +1,100 @@
...
+
+* Community members - participants of our community.
+* Administrators - members who own and administer the project.
+
+## Community members
+
+### Team
+
+Anybody who contributes code to one of our project is considered a community
+member.
```
zowoq, line 22: I don't see how this definition is feasible if community members have some sort of voting rights?

tomberek: Generally, open-source authority stems from a combination of several factors; contribution, longevity, and dependability.

1) Have they contributed value?
2) Have they shown a long-term interest in the project?
3) Have they been responsive to the needs of the project?

Only basing this on any single of the above factors leads to odd situations.

docs/rulebook.md
adisbladis: Not too much on this repo any more. I'd still like to be an org admin regardless as I do way more in that role than I do in this repo.

docs/rulebook.md
zimbatm: There are a lot of dimensions that make a good contributor. Sometimes the user only hangs out on Matrix and helps users out. Or they submitted one patch that has an incredible impact. Or they help maintain a good spirit in the community just by who they are. Or ...

Reversely, some of the value of contribution may decay over time. Should users that submitted a patch 10 years ago still have voting rights?

It's hard to find a single metric that can be used as an inclusion criterion. And those metrics can probably be gamed pretty easily.

I think what's important is that you can feel like you have a voice when participating in this community. And that you can be confident that the project is handled well.

zimbatm: How about the following criteria:

* Has at least 10 contributions to projects that are in the org
* The last contribution is younger than 2 years old

docs/rulebook.md
```diff
@@ -0,0 +1,100 @@
...
+
+To reduce the attack surface on the project, the administrative team size is
+limited to 5 people.
+
+### Power
+
+The administrators have the power to change any aspect of the project.
+
+Controversial changes require a majority vote. Any administrator can call a
+change controversial :-)
```
zimbatm, line 76: TODO: Either change the term "controversial" or clarify that "controversial" is not a negative term in this context. It just means that opinions differ and we need a bit more time to reach a consensus.

tomberek (approved):

docs/rulebook.md
zimbatm: fixed in ba214bd0cfca7ab5e2df4a61a0c1b4de0446e2bb

zowoq:
I'd like to move forward with this but honestly I don't like the current form with FIXMEs and brief statements about members, voting, etc.

Formalizing what seems to be the status quo (admin team responsibilities, quorum, etc) is fairly simple, I'd like to leave the "new" and less simple stuff (defining "community members", referendums, voting and so forth) to another PR.

zowoq:
Is a board of elected (annually ?) community members something that may be feasible instead of referendums?

zimbatm:
Sounds good. It's better to move slowly and not add too many "laws". It was a good idea to extract the MVP as a separate PR. So maybe we can continue doing more of this.

What matters to me is that this project should always belong to the community first. People are trusting us to hold and respect the work they put into their repos. We (the admins) are here to facilitate their work.

On top of the rules, it would also be nice to describe common practices and approaches. For example, what happens if a maintainer disappears?

zimbatm:
Instead of defining rules or a constitution, let's document how we operate. What matters is to communicate to other people how we work, so that they know how to participate in the project.

PR #385, opened 2023-01-20
.github/settings.yml: fix mergify

zowoq:


Mic92:
We can actually get rid of mergify now, by having our update github actions just include a `bors merge` in their message, no?

zowoq:
`bors merge` only works if you have write access to the repo so it won't work with PRs that are opened by `github-actions`.

Guess we could give a bot account access to this repo and use it's token in the action to open PRs?

PR #389, opened 2023-01-21
sops: remove GPG

zimbatm:
- direnv: also reload on shell.nix and treefmt.nix changes
- sops: replace my GPG key with AGE


zimbatm:
After this change I'm still getting this error when entering the shell:
```
environment:2719: sopsCreateGPGHome: unbound variable

```

It disappears if I comment out `sops-import-keys-hook` in shell.nix.

Mic92:
bors merge

PR #403, opened 2023-01-23
terraform/terraform wrapper

zimbatm:
Renamed the wrapper because all the usages of ./terraform/deploy were
invoking terraform. I also fixed an escaping issue that I encountered
when running `./terraform import 'some.resource["xxx"]' ID`


zowoq:
bors merge

zowoq:
Might be able to just drop the wrapper altogether if we switch to using the sops provider?

zimbatm:
Yes, that would work too. It would allow us to be more explicit about who needs what secret as well.

Issue #388, opened 2023-01-21
Enable discussions on the org?


zimbatm:
Should we try to enable the discussions? This might help create a more community feel to the org. Let people raise issues, and have longer form communication than on the Matrix channel.

![image](https://user-images.githubusercontent.com/3248/213860793-214d2e2f-e4c2-40b1-bbf2-4fcc3f077aae.png)

zowoq:
Seems like a good idea, less friction to use and doesn't fragment or obscure discussion like running another separate forum would.

👍

zimbatm:
Ok, we can try it and see how it goes

zimbatm:
I'm also proposing that we announce infrastructure changes in the forum, as per:
https://github.com/orgs/nix-community/discussions/401

PR #405, opened 2023-01-23
docs: list all the support options

zimbatm:
Part of #388

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -35,11 +35,21 @@ The maintainer of a project gets admin access to the project they maintain. The 
 
 If you see a project that is under-maintained, submit PRs to fix it as if it was your own project. If the PRs don't get merged, ping the [nix-community administrators](#nix-community-administrators).
 
+## Support
+
+For casual chat, come join us in the [nix-community](https://matrix.to/#/#nix-community:nixos.org) room.
+
+For urgent issues, ping us in the Matrix room (see the admin list below).
+
+For repo additions or service requests, create an issue here:
+[New Issue](https://github.com/nix-community/infra/issues/new).
+
+For general conversation about governance and direction of the project, use
+the [Discussion forum](https://github.com/orgs/nix-community/discussions).
+
 ## nix-community administrators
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -39,7 +39,17 @@ If you see a project that is under-maintained, submit PRs to fix it as if it was
 
 If you need to contact the nix-community admins, they are listed [here](https://github.com/nix-community/infra#nix-community-administrators).
 
-Come chat with us in [#nix-community](https://matrix.to/#/#nix-community:nixos.org) using the matrix protocol!
+## Support
+
```
zowoq, line 43: Could we just have a link to https://github.com/nix-community/infra#support here like we do for the admin list above?

zimbatm: I thought about it but it adds friction to people coming on the website. We probably want to consolidate the site, README and also the .github/README at some point.

zowoq: > I thought about it but it adds friction to people coming on the website.

If they want a list of admins they need to click a link anyway so I don't think it's any worse.

docs/README.md
```diff
@@ -39,7 +39,17 @@ If you see a project that is under-maintained, submit PRs to fix it as if it was
...
+For casual chat, come join us in the [nix-community](https://matrix.to/#/#nix-community:nixos.org) room.
+
+For urgent issues, ping us in the Matrix room (see the admin list below).
```
zowoq, line 46: To keep these sections the same probably easier to move support above admins.

docs/README.md
zimbatm: This is only true for 1/4 of contact methods. It's important that visitors see that info at a first glance and not have to go dig for it.

But also, we probably want to refactor all of this. I will merge this as it's already an improvement and keep iterating on the docs.

docs/README.md
zimbatm: thanks

zimbatm:
bors merge

PR #415, opened 2023-01-24
terraform: use sops provider, drop wrapper

zowoq:

### DIFF
--- a/.github/workflows/terraform.yml
+++ b/.github/workflows/terraform.yml
@@ -18,6 +18,7 @@ jobs:
       SOPS_AGE_KEY: ${{ secrets.SOPS_AGE_KEY }}
       TF_TOKEN_app_terraform_io: ${{ secrets.TF_TOKEN_APP_TERRAFORM_IO }}
       TF_IN_AUTOMATION: true
+      TF_INPUT: 0
     steps:
       - uses: actions/checkout@v3
       - uses: dorny/paths-filter@v2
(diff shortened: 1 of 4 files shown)


Mic92 (approved):
Using this in other places. Works great.

zowoq:
Still need a way to set the backend token ...

zimbatm:
Now that all the admins have access to the backend, it's probably best to use individual tokens there

.github/workflows/terraform.yml
```diff
@@ -11,8 +11,12 @@ jobs:
   terraform-deploy:
     if: github.repository == 'nix-community/infra'
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: terraform
     env:
       SOPS_AGE_KEY: ${{ secrets.SOPS_AGE_KEY }}
+      TF_TOKEN_app_terraform_io: ${{ secrets.TF_TOKEN_APP_TERRAFORM_IO }}
```
zowoq, line 19: New team api token.

terraform/terraform
```diff
@@ -1,9 +0,0 @@
-#!/usr/bin/env bash
-set -euo pipefail
-
-pushd "$(dirname "$0")" >/dev/null
-
-# terraform cloud without the remote execution part
-TF_FORCE_LOCAL_BACKEND="1" sops exec-env secrets.yaml "terraform ${*@Q}"
```
zowoq, line 7: I don't think we still need `TF_FORCE_LOCAL_BACKEND` but not really sure, there isn't much information about it

zowoq:
I'll leave the sops secrets as is for now so we don't need to deal with a merge conflict there.

terraform/terraform
zimbatm: This isn't needed anymore after 
https://github.com/nix-community/infra/blob/9d68769b4326999bd114b4c08fe0a39b89a49709/terraform/terraform_cloud.tf#L46

zimbatm:
Let's merge this before the gandi changes

zowoq:
bors merge

PR #418, opened 2023-01-25
use gandi for our domains

zimbatm:

### DIFF
--- a/terraform/gandi.tf
+++ b/terraform/gandi.tf
@@ -1,0 +1,11 @@
+# FIXME: Not declared because needs the owners block and so would require us
+# to expose personal mail addresses to the public.
+#
+# resource "gandi_domain" "nix_community" {
+#   name = "nix-community.org"
+# }
+
+resource "gandi_email_forwarding" "admin" {
+  source       = "admin@nix-community.org"
+  destinations = values(local.admins)
+}
--- a/terraform/terraform_providers.tf
+++ b/terraform/terraform_providers.tf
@@ -3,6 +3,9 @@ terraform {
     cloudflare = {
       source = "cloudflare/cloudflare"
     }
+    gandi = {
+      source = "go-gandi/gandi"
+    }
     hydra = {
       source = "DeterminateSystems/hydra"
     }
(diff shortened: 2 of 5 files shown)


zowoq:
Opened an alternative for the sponsors: https://github.com/nix-community/infra/pull/419

terraform/terraform_providers.tf
```diff
@@ -18,3 +21,5 @@ provider "hydra" {
   host     = "https://hydra.nix-community.org"
   username = "admin"
 }
+
+provider "gandi" {}
```
zowoq, line 25: Could we move this above hydra please?

terraform/gandi.tf
```diff
@@ -0,0 +1,11 @@
+# FIXME: Not declared because needs the owners block and so would require us
```
zowoq, line 1: Would storing it in sops and using the `obfuscated` options work?

zowoq: Actually maybe we just don't bother with this? Does require a lot of info.

terraform/terraform_providers.tf
zowoq: Moved this and fixed the docs merge conflict.

zowoq:
```
gandi_email_forwarding.admin: Creating...
╷
│ Error: StatusCode: 404 ; Err: 404: The resource could not be found.
│
│   with gandi_email_forwarding.admin,
│   on gandi.tf line 8, in resource "gandi_email_forwarding" "admin":
│    8: resource "gandi_email_forwarding" "admin" {
│
╵
exit status 1
```
Not sure what the problem is.

terraform/gandi.tf
nlewo: Yep, i think this would add unnecessary complexity.

Since the domain is already created, you don't need to create it and  you could not delete the domain since this is not allowed by the public API :/

zimbatm:
Blocked on https://github.com/go-gandi/terraform-provider-gandi/pull/137

zimbatm:
bors try

zimbatm:
bors try

zimbatm:
bors merge

PR #427, opened 2023-02-01
terraform: pin terraform and providers in separate flake

zowoq:

### DIFF
--- a/ci.nix
+++ b/ci.nix
@@ -4,6 +4,7 @@
 }:
 let
   self = builtins.getFlake (toString ./.);
+  terraform = builtins.getFlake (toString ./terraform/.);
   inherit (self.inputs) nixpkgs;
   stripDomain = name: nixpkgs.lib.head (builtins.match "(.*).nix-community.org" name);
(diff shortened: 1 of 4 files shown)


zowoq:
bors try

zowoq:
bors try

zowoq:
Think we might want to split the terraform shell out so we can pin it independent of the system updates.

zowoq:
bors try

zimbatm:
Is terraform blocking the system updates?

zowoq:
No, this is to avoid any potential issues with the providers being constantly updated so we can use the terraform action from a known configuration.

zowoq:
Rebased to pick up the gandi changes and dropped the commit merging shell.nix into the flake, doesn't really make a difference if it's a separate file or not.

zowoq:
Cloudflare provider has been bumped to v4 which has breaking changes so I'll merge this in a day or two before the next flake update.

zowoq:
bors try

PR #429, opened 2023-02-03
users: remove admins after quorum changes

zowoq:
moved to roles/builder/users as trusted to keep access to the community builder

follow up to 232ff1eee52e01fe8732fab281f14db54ed2c1de

cc @nix-community/infra


zimbatm (approved):
Fair

PR #432, opened 2023-02-06
flake: drop domain

zowoq:

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -5,7 +5,7 @@ import os
 import subprocess
 import sys
 from pathlib import Path
-from typing import Any, List
+from typing import List
 
 from deploykit import DeployGroup, DeployHost
 from invoke import task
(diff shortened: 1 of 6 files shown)


tasks.py
```diff
@@ -1,101 +1,38 @@
...
-
-
-def _format_disks(host: DeployHost, devices: List[str]) -> None:
-    assert (
-        len(devices) == 1 or len(devices) == 2
-    ), "we only support single devices or mirror raids at the moment"
-    # format disk with as follow:
-    # - partition 1 will be the boot partition, needed for legacy (BIOS) boot
-    # - partition 2 is for boot partition
-    # - partition 3 takes up the rest of the space and is for the system
```
Mic92, line 52: Maybe it would be good to already write some disko configuration since we otherwise don't have anything to document the current disk layout.

Mic92: Or at least copy out the current code into an issue with a TODO to fix it eventually.

zowoq: I'll take a look at writing the disko config.

zowoq: I've kept a copy of the current config in https://github.com/nix-community/infra/pull/485.

zowoq:
As the hostnames are changing we'll need to deploy this manually.

tasks.py
```diff
@@ -28,74 +28,25 @@ def deploy_nixos(hosts: List[DeployHost]) -> None:
...
+        h.run(
+            [
+                "sudo",
+                "nixos-rebuild",
+                "switch",
+                "--option",
+                "accept-flake-config",
+                "true",
+                "--flake",
+                f".#{hostname}",
```
Mic92, line 45:
```suggestion
                f"/etc/nixos#{hostname}",
```

Mic92:
bors merge

tasks.py
```diff
@@ -218,56 +169,11 @@ def update_terraform(c):
...
-def format_disks(c, hosts="", disks=""):
-    """
-    Format disks with zfs, i.e.: inv format-disks --hosts build02 --disks /dev/nvme0n1,/dev/nvme1n1
-    """
-    for h in get_hosts(hosts):
-        _format_disks(h, disks.split(","))
-
-
-@task
-def setup_secret(c, hosts=""):
```
Mic92, line 231: Nixos-anywhere also has an option to provision secrets which would come handy for deploying ssh host keys.

Mic92:
@nix-community/admin If you want to deploy you now need to add this to your ssh config if your username on the server is different from your local username on your own workstation:

```
Host *.nix-community.org
  User <youradminusername>
```

i.e.:

```
Host *.nix-community.org
  User mic92
```

zowoq:
> If you want to deploy you now need to add this to your ssh config if your username on the server is different from your local username on your own workstation

I'll add this to the docs.

tasks.py
zowoq: Thanks for catching that.

PR #438, opened 2023-02-16
terraform: use Gandi for email

zimbatm:
Fix the DNS records so we can use the email forwarding service from
Gandi.


zowoq (approved):
treefmt is unhappy

ryantm (approved):

zimbatm:
bors merge

PR #453, opened 2023-02-24
terraform/hydra: add jobset for go 1.20 bump

zowoq:
cc @Mic92 @zimbatm 

I think this is correct? Want to check x86_64-linux nixosTests.


Mic92:
Not so sure if our builder can handle this. It might run out of disk space. Can you not ask @vcuncat to get a jobset on the public hydra?

PR #455, opened 2023-02-27
roles/hercules-ci: try workaround for "could not retrieve derivation"

zowoq:


zowoq:
Doesn't make a difference.

PR #457, opened 2023-03-01
tasks.py: drop rsync, run nixos-rebuild remotely

zowoq:

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -9,15 +9,6 @@ from typing import Any, List
 from deploykit import DeployGroup, DeployHost
 from invoke import task
 
-RSYNC_EXCLUDES = [
-    ".direnv",
-    ".git",
-    ".mypy-cache",
-    ".ruff_cache",
-    ".terraform",
-    "result*",
-]
-
 
 def deploy_nixos(hosts: List[DeployHost]) -> None:
     """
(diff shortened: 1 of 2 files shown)


tasks.py
```diff
@@ -26,13 +17,23 @@ def deploy_nixos(hosts: List[DeployHost]) -> None:
     g = DeployGroup(hosts)
 
     def deploy(h: DeployHost) -> None:
-        target = f"{h.user or 'root'}@{h.host}"
         h.run_local(
-            f"rsync {' --exclude '.join([''] + RSYNC_EXCLUDES)} -vaF --delete -e ssh . {target}:/etc/nixos"
```
Mic92, line 31: I did this mainly for speed because the remote machines is at least as powerful as my local machine and this way I don't have to upload packages to the remote builder.
Can you elaborate on why you want to switch to `nixos-rebuild`?

tasks.py
```diff
@@ -26,13 +17,23 @@ def deploy_nixos(hosts: List[DeployHost]) -> None:
...
+                "nixos-rebuild",
+                "switch",
+                "--option",
+                "accept-flake-config",
+                "true",
+                "--flake",
+                f".#{h.host}",
+                "--fast",
+                "--build-host",
+                f"{h.host}",
```
Mic92, line 31: If you set a build host it will cause all derivations to be uploaded. This is useful for cross-arch deployments but otherwise it's significant slower than a normal nixos-rebuild.

```suggestion
                f"",
```

tasks.py
zowoq: Want to switch away from rsync and needing to keep an exclude list for it, could switch to something else beside nixos-rebuild?

tasks.py
zowoq: Wouldn't we need this for x86_64-linux -> aarch64-linux or darwin -> linux deployments?

tasks.py
Mic92: How about this alternative? That wouldn't need maintaining a filter.

```
rsync "$(nix flake info --json | jq -r .path)" ...
```

Mic92: How do you like this? https://github.com/nix-community/infra/pull/460

zowoq: That's good, thank you.

PR #460, opened 2023-03-06
change deployment to use flake only

Mic92:

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -9,15 +9,6 @@ from typing import Any, List
 from deploykit import DeployGroup, DeployHost
 from invoke import task
 
-RSYNC_EXCLUDES = [
-    ".direnv",
-    ".git",
-    ".mypy-cache",
-    ".ruff_cache",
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -9,26 +9,21 @@
...
-]
-
 
 def deploy_nixos(hosts: List[DeployHost]) -> None:
     """
     Deploy to all hosts in parallel
     """
     g = DeployGroup(hosts)
 
+    res = subprocess.run(["nix", "flake", "info", "--json"], check=True, text=True, stdout=subprocess.PIPE)
```
zowoq, line 19: > warning: 'nix flake info' is a deprecated alias for 'nix flake metadata'

```suggestion
    res = subprocess.run(["nix", "flake", "metadata", "--json"], check=True, text=True, stdout=subprocess.PIPE)
```

zowoq (commented):
treefmt check is failing

PR #464, opened 2023-03-09
build04: switch to systemd-boot

Mic92:
cc @zowoq
I was greeted by a EDK2 EFI shell and had to boot grub manually.
Systemd-boot meanwhile seems to work fine and also installs a default boot loader path that would work even if efivars is not set.
bors merge


Mic92:
bors merge

PR #465, opened 2023-03-10
add update hound action

zowoq:

### DIFF
--- a/.github/workflows/hound.yml
+++ b/.github/workflows/hound.yml
@@ -1,0 +1,38 @@
+name: "Update hound"
+
+on:
+  repository_dispatch:
+  workflow_dispatch:
+  schedule:
+    - cron: "51 2 * * 3"
+
+jobs:
+  createPullRequest:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Nix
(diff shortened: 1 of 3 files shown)


.github/workflows/hound.yml
```diff
@@ -0,0 +1,37 @@
...
+            experimental-features = nix-command flakes
+      - name: Setup
+        id: setup
+        run: |
+          echo "title=hound: update repos $(date -u +"%Y-%m-%d")" >> $GITHUB_OUTPUT
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config user.name "github-actions[bot]"
+      - name: Update hound repos
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN_FOR_UPDATES }}
```
Mic92, line 27: Token should be fine, potentially not required, you can also trigger CI using bors and our hercules one should be triggered also with the github bot provided tokens.

zowoq: This is to avoid the repo update script being rate limited. 

Default token would work, I'll change it for both.

Mic92: Upside of that is that we can also see if user manually triggered PR creation this way.

Mic92:
bors merge

PR #470, opened 2023-03-12
tasks.py: merge update-hound.py script

zowoq:
I forgot about channels / nix_path, the hound github action won't work.

Rather than adding them to the action I've merged the update-hound.py script into tasks.py.


tasks.py
```diff
@@ -10,10 +10,9 @@
 from invoke import task
```
Mic92, line 12: since invoke can be also triggered from a subdirectory:

```
from pathlib import Path

ROOT = Path(__file__).parent.resolve()
os.chdir(ROOT)
```

zowoq: Added.

PR #474, opened 2023-03-13
treefmt

zowoq:
Terraform is only used in this flake for formatting so may as well switch to a lighter alternative.


zimbatm (approved):
bors merge

PR #485, opened 2023-03-18
preparation for switching to disko

zowoq:
```py
def sfdisk_json(host: DeployHost, dev: str) -> List[Any]:
    out = host.run(f"sfdisk --json {dev}", stdout=subprocess.PIPE)
    data = json.loads(out.stdout)
    return data["partitiontable"]["partitions"]


def _format_disks(host: DeployHost, devices: List[str]) -> None:
    assert (
        len(devices) == 1 or len(devices) == 2
    ), "we only support single devices or mirror raids at the moment"
    # format disk with as follow:
    # - partition 1 will be the boot partition, needed for legacy (BIOS) boot
    # - partition 2 is for boot partition
    # - partition 3 takes up the rest of the space and is for the system
    for device in devices:
        host.run(
            f"sgdisk -Z -n 1:2048:4095 -n 2:4096:+2G -N 3 -t 1:ef02 -t 2:8304 -t 3:8304 {device}"
        )

    # create mdadm raid for /boot with ext4
    if len(devices) == 2:
        boot_parts = []
        root_parts = []
        for dev in devices:
            # use partuuids as they are more stable than device names
            partitions = sfdisk_json(host, dev)
            boot_parts.append(partitions[1]["node"])
            root_parts.append(f"/dev/disk/by-partuuid/{partitions[2]['uuid'].lower()}")

        host.run(
            f"mdadm --create --verbose /dev/md127 --raid-devices=2 --level=1 {' '.join(boot_parts)}"
        )
        host.run(
            f"zpool create zroot -O acltype=posixacl -O xattr=sa -O compression=lz4 mirror {' '.join(root_parts)}"
        )
        boot = "/dev/md127"
    else:
        partitions = sfdisk_json(host, devices[0])
        boot = partitions[1]["node"]
        uuid = partitions[2]["uuid"].lower()
        root_part = f"/dev/disk/by-partuuid/{uuid}"
        host.run(
            f"zpool create zroot -O acltype=posixacl -O xattr=sa -O compression=lz4 -O atime=off {root_part}"
        )

    host.run("partprobe")
    host.run(f"mkfs.ext4 -F {boot}")

    # setup zfs dataset
    host.run("zfs create -o mountpoint=none zroot/root")
    host.run("zfs create -o mountpoint=legacy zroot/root/nixos")
    host.run("zfs create -o mountpoint=legacy zroot/root/home")

    ## and finally mount
    host.run("mount -t zfs zroot/root/nixos /mnt")
    host.run("mkdir /mnt/home /mnt/boot")
    host.run("mount -t zfs zroot/root/home /mnt/home")
    host.run("mount -t ext4 /dev/md127 /mnt/boot")


@task
def format_disks(c, hosts="", disks=""):
    """
    Format disks with zfs, i.e.: inv format-disks --hosts build02 --disks /dev/nvme0n1,/dev/nvme1n1
    """
    for h in get_hosts(hosts):
        _format_disks(h, disks.split(","))
```

### DIFF
--- a/roles/disko.nix
+++ b/roles/disko.nix
@@ -1,0 +1,98 @@
+{ disks ? [ ], ... }:
+let
+  content = {
+    format = "gpt";
+    type = "table";
+    partitions = [
+      {
+        name = "boot";
+        start = "0";
+        end = "1M";
+        part-type = "primary";
+        flags = [ "bios_grub" ];
+      }
+      {
+        name = "mdadm";
+        start = "1M";
+        end = "2G";
+        content = {
+          name = "boot";
+          type = "mdraid";
+        };
+      }
+      {
+        name = "zfs";
+        start = "2G";
+        end = "-16G";
+        content = {
+          pool = "zroot";
+          type = "zfs";
+        };
+      }
+      {
+        name = "swap";
+        start = "-16G";
+        end = "100%";
+        part-type = "primary";
+        content = {
+          type = "swap";
+        };
+      }
+    ];
+  };
+in
+{
+  disk = {
+    disk0 = {
+      device = builtins.elemAt disks 0;
+      type = "disk";
+      inherit content;
+    };
+    disk1 = {
+      device = builtins.elemAt disks 1;
+      type = "disk";
+      inherit content;
+    };
+  };
+  mdadm = {
+    boot = {
+      type = "mdadm";
+      level = 1;
+      metadata = "1.0";
+      content = {
+        type = "filesystem";
+        format = "ext4";
+        mountOptions = [ "nofail" ];
+        mountpoint = "/boot";
+      };
+    };
+  };
+  zpool = {
+    zroot = {
+      type = "zpool";
+      rootFsOptions = {
+        acltype = "posixacl";
+        atime = "off";
+        compression = "lz4";
+        xattr = "sa";
+        "com.sun:auto-snapshot" = "false";
+      };
+
+      datasets = {
+        "root" = {
+          type = "zfs_fs";
+          mountpoint = null;
+        };
+        "root/nixos" = {
+          type = "zfs_fs";
+          mountpoint = "/";
+        };
+        "root/home" = {
+          type = "zfs_fs";
+          mountpoint = "/home";
+          mountOptions = [ "nofail" ];
+        };
+      };
+    };
+  };
+}
--- a/build01/configuration.nix
+++ b/build01/configuration.nix
@@ -12,7 +12,8 @@
 {
   imports = [
     ../roles/common.nix
-    ../roles/zfs-raid.nix
+    ../roles/raid.nix
+    ../roles/zfs.nix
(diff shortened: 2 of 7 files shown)


roles/disko.nix
```diff
@@ -0,0 +1,49 @@
...
+let
+  content = {
+    format = "gpt";
+    type = "table";
+    partitions = [ ];
+  };
+in
+{
+  disk = {
+    disk0 = {
```
Mic92, line 11: I was wondering if we should go for a mirror raid since we wear out the NVME quite a bit with write cycles and we don't really need that high of an uptime.

zowoq: Did you mean striping instead of mirroring?

We probably want to have a swap partition as well.

Mic92: Ah. Yes. I mean striping instead of mirroring.

roles/disko.nix
```diff
@@ -0,0 +1,95 @@
...
+      metadata = "1.0";
+      content = {
+        type = "filesystem";
+        format = "ext4";
+        mountOptions = [ "nofail" ];
+        mountpoint = "/boot";
+      };
+    };
+  };
+  zpool = {
```
zowoq, line 70: `zpool` is a copy of the build04 config with the addition of `mountOptions = [ "nofail" ];` for `root/home`.

Not very familiar with zfs so not sure we still need to set acltype xattr atime or legacy mountpoints like we did in the original disk config?

roles/disko.nix
```diff
@@ -0,0 +1,95 @@
...
+    zroot = {
+      type = "zpool";
+      rootFsOptions = {
+        compression = "lz4";
```
zowoq, line 74: Seems zstd is an option here but not sure if we should use it?

zowoq (commented):
The disko config is mostly copy pasted from the upstream examples so I may have missed something.

roles/disko.nix
Mic92: You can keep all the options as they are best practices. You would need to configure either legacy mountpoints or use the zfsutil mount option so that systemd is allowed to mount zfs filesystems: https://github.com/Mic92/dotfiles/blob/b092cce7cc0e3c84dccf84e69f783fceeef09e79/nixos/blob64/hardware-configuration.nix#L14

roles/disko.nix
Mic92: With nvme strip raid this might be a bottleneck. Decent choice for harddrives though.

roles/disko.nix
zowoq: Added the options, `zfsutil` seems to be the default that is set by disko.

zowoq:
@Mic92 Now we have backups of the nixpkgs-update logs I'm planning merging this and redeploying build01,build02,build03 in the coming week.

zowoq:
I'll split this up so it's only one machine at a time.

PR #489, opened 2023-03-20
flake: drop unused sops-nix input

zowoq:


flake.nix
```diff
@@ -69,7 +70,7 @@
             common = [
               { _module.args.inputs = inputs; }
               { srvos.flake = inputs.self; }
-              inputs.sops-nix.nixosModules.sops
+              inputs.sops-nix.nixosModules.default
```
zowoq, line 73: Updated but still has the same warning?

> https://hercules-ci.com/github/nix-community/infra/jobs/1136
> trace: warning: use sops.nixosModules.default instead of sops.nixosModule

PR #497, opened 2023-03-27
.github/workflows/flake-updates.yml: remove automatic merge

zowoq:
For some reason the update PRs have been really flaky recently. Including `bors merge` in the PR kicks of two builds at the same time, I'd disable that and see if the issue still persists. I'm usually around when these PRs are opened so there won't be much delay by not merging it automatically.


Mic92:
bors merge

PR #506, opened 2023-04-07
build02/nixpkgs-update: enable another worker

zowoq:
@mic92 The oom's seem to have settled down, shall we try adding another worker?


Mic92:
Maybe @ryantm is better to answer the question I don't know how much the application benefits from more workers since it will also issues build that are parallel.

ryantm (approved):
Sure let's try

zowoq:
Reverted this as it was using too much memory.

PR #511, opened 2023-04-11
build03: disable hydra

zowoq:
post-build-hook is failing and blocking builds that are scheduled on build03

### DIFF
--- a/build03/configuration.nix
+++ b/build03/configuration.nix
@@ -16,7 +16,7 @@
     ../roles/remote-builder/aarch64-build04.nix
 
     ../services/hound
-    ../services/hydra
+    #../services/hydra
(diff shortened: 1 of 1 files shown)


Mic92:
bors merge

zowoq:
bors merge

zowoq:
@nlewo 
@nix-community/emacs 

Temporarily disabling hydra as the post-build-hook is failing and blocking other builds. We have a couple of alternatives for the hook so I don't expect that it disabled for more than a day or two.

zowoq:
Hydra is enabled and working its way through the queue. Seems it was a problem with cachix itself so I've pinned it to an earlier version for now.

PR #513, opened 2023-04-11
remove post-build-hook, switch to cachix watch-store

zowoq:
watch-store seems to be more efficient than the hook.

https://github.com/cachix/cachix/blob/master/cachix/CHANGELOG.md#13---2023-03-06

with >=1.3 it defaults to 8 jobs

### DIFF
--- a/services/hydra/default.nix
+++ b/services/hydra/default.nix
@@ -3,21 +3,6 @@ with lib; let
   cfg = config;
 
   hydraPort = 3000;
-
-  upload-to-cachix = pkgs.writeScriptBin "upload-to-cachix" ''
-    #!/bin/sh
-    set -eu
-    set -f # disable globbing
-
-    # skip push if the declarative job spec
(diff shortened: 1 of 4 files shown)


zowoq:
Same problem with this as well.

```
Warning: /nix/store/*** is not valid, skipping
cachix: FatalError {fatalErrorMessage = "no such path /nix/store/***"}
```

Mic92 (approved):

services/hydra/default.nix
```diff
@@ -43,11 +28,10 @@ in
     # hydra-queue-runner needs to read this key for remote building
     sops.secrets.id_buildfarm.owner = "hydra-queue-runner";
 
-    nix.extraOptions = ''
-      builders-use-substitutes = true
```
zimbatm, line 47: This one disappeared, but it's in srvos so it's OK

zimbatm (approved):

PR #516, opened 2023-04-12
flake: add nix-community/flake-compat

zowoq:
It's easier to be able to fall back to the old cli sometimes.

I'm fine with not adding it if anyone is 👎 .

cc @nix-community/admin

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,16 @@
+# This file provides backward compatibility to nix < 2.4 clients
+{ system ? builtins.currentSystem }:
+let
+  lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+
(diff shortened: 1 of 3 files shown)


Mic92 (approved):
Don't use it but also don't object.

zimbatm (approved):
sure :+1:

zowoq:
I should be able to use a couple of local workarounds instead so I don't really need this.

zowoq:
One year later I'm still occasionally find that having flake-compat would be helpful.

@zimbatm Is https://github.com/nix-community/flake-compat still maintained or should I switch to https://github.com/edolstra/flake-compat?

zimbatm:
Not actively maintained, but still easier to get patches merged in

PR #517, opened 2023-04-12
treefmt: add deadnix

zowoq:
This mostly avoids my original objection to having deadnix enabled as now it isn't all or nothing. We can keep dead/commented out code without needing to disable the other nix formatters.

I think it's possible that the deadnix and nix formatters could conflict but that doesn't seem like too much of a hassle.

### DIFF
--- a/treefmt.nix
+++ b/treefmt.nix
@@ -12,6 +12,8 @@
       options = [
         "-eucx"
         ''
+          ${pkgs.lib.getExe pkgs.deadnix} --edit "$@"
+
           for i in "$@"; do
             ${pkgs.lib.getExe pkgs.statix} fix "$i"
           done


treefmt.nix
```diff
@@ -1,14 +1,22 @@
 {
-  perSystem = { pkgs, ... }: {
+  perSystem = { config, pkgs, ... }: {
     treefmt = {
       # Used to find the project root
       projectRootFile = "flake.lock";
 
+      programs.deadnix.enable = true;
```
Mic92, line 7: afaik treefmt only supports one formatter per extension. so I don't think enabling this option will have any effect or it will disable existing formatters below depending on the order.

zimbatm: technically it works, but both formatters might touch the same file at the same time

Mic92: sounds racy.

zimbatm: in practice I never had a problem, but yeah

zowoq: The proper (and configurable) fix would be sequences which has an open issue in treefmt but could a "cheaper" fix be making the order of formatters deterministic and doing something like `a_deadnix` `b_nixpkgs`?

zowoq: That doesn't allow separate excludes for deadnix which is what I was trying for with this approach.

zowoq: I guess adding a ignore directive to deadnix would also work and would allow us to have deadnix, statix and nixpkgs-fmt all under the nix formatter.

zowoq: https://github.com/astro/deadnix/commit/03b7f9a84086c58be91245a589c8c9144ffa6e2e

Upsteam added `#deadnix: skip` so we can have deadnix under `nix` once it lands in nixpkgs.

PR #518, opened 2023-04-12
treefmt: flatten the config

zimbatm:
By importing the config on the treefmt module level, it makes the config more flat, and also compatible with non-flake-parts users.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -60,8 +60,9 @@
           inputs.treefmt-nix.flakeModule
           ./effect.nix
           ./shell.nix
-          ./treefmt.nix
         ];
+
+        perSystem.treefmt.imports = [ ./treefmt.nix ];
 
         flake.nixosConfigurations =
           let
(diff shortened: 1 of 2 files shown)


zowoq:
> By importing the config on the treefmt module level, it makes the config more flat

I kind of just undid this for shell.nix in c0bc25fab37278c63b0867e88efd306e7190b7c4 so everything was in imports and (mostly) consistent, I guess it could be changed back?

> also compatible with non-flake-parts users.

Afraid I don't understand this?

zimbatm:
> Afraid I don't understand this?

It's compatible with `treefmt-nix.lib.mkWrapper nixpkgs ./treefmt.nix` for example.

Overall I'm quite fond of putting the treefmt config in this `treefmt.nix` file as a convention since it's replacing the `treefmt.toml` file, and this is already the case.

There is tension between making all the modules flake-part compatible or not as you noted; since you're the one maintaining the repo, I will let you decide if you want to merge this PR or not.

Mic92 (approved):
Looks simpler.

PR #520, opened 2023-04-14
hydra

zowoq:

### DIFF
--- a/services/hydra/default.nix
+++ b/services/hydra/default.nix
@@ -1,18 +1,13 @@
 { lib, pkgs, config, ... }:
-with lib; let
-  cfg = config;
-
-  hydraPort = 3000;
-in
 {
   options.services.hydra = {
-    adminPasswordFile = mkOption {
-      type = types.str;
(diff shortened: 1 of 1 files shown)


services/hydra/default.nix
```diff
@@ -33,18 +28,6 @@ in
       "https://github.com/NixOS/"
     ];
 
-    nixpkgs.config = {
```
zowoq, line 36: @zimbatm nixpkgs.config was added in 19581331ed38654514ee14f09a5c19123da7da8e, I don't think we need it now?

zimbatm: I don't know if those still apply now that we use `nixpkgs.lib.nixosSystem`? I think it's using the passed instance of nixpkgs instead.

zowoq: Sorry, I meant that I don't think that we're still running builds for anything that requires them?

zimbatm: I propose merging this and find out :)

PR #521, opened 2023-04-15
add terrafrom/github-to-gitlab-mirror

Mic92:

### DIFF
--- a/terraform/github-to-gitlab-mirror/.envrc.private-template
+++ b/terraform/github-to-gitlab-mirror/.envrc.private-template
@@ -1,0 +1,2 @@
+# Got to https://gitlab.com/-/profile/personal_access_tokens
+export GITLAB_TOKEN=<your_token>
--- a/terraform/github-to-gitlab-mirror/get-last-commit.sh
+++ b/terraform/github-to-gitlab-mirror/get-last-commit.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+if [ -z "$1" ]; then
+  echo "Usage: $0 github_repo"
+  exit 1
+fi
+github_repo=$1
+last_commit=$(git ls-remote "https://github.com/$github_repo" HEAD | awk '{ print $1}')
+# echo as json
+echo "{\"last_commit\": \"$last_commit\"}"
--- a/terraform/github-to-gitlab-mirror/main.tf
+++ b/terraform/github-to-gitlab-mirror/main.tf
@@ -1,0 +1,59 @@
+locals {
+  user = "nix-community"
+  blacklist = [
+    # invalid name
+    ".github",
+    # do already exists
+    "emacs-overlay",
+    "docker-nixpkgs",
+    "nixos-generators",
+    "poetry2nix",
+  ]
+  # remove blacklist
+  filteredRepos = [
+    for repo in data.github_repositories.repos.full_names : repo if !contains(local.blacklist, element(split("/", repo), 1))
+  ]
+}
+
+data "github_repositories" "repos" {
+  query = "user:${local.user} is:public"
+}
+
+data "github_repositories" "non-archived-repos" {
+  query = "user:${local.user} is:public archived:false"
+}
+
+resource "gitlab_group" "group" {
+  name             = local.user
+  path             = local.user
+  visibility_level = "public"
+}
+
+resource "gitlab_project" "repos" {
+  for_each               = toset(local.filteredRepos)
+  name                   = element(split("/", each.key), 1)
+  namespace_id           = gitlab_group.group.id
+  import_url             = "https://github.com/${each.key}"
+  shared_runners_enabled = false
+  visibility_level       = "public"
+}
+
+# Cannot use api here or we run into api rate limits.
+data "external" "repos" {
+  for_each = toset(local.filteredRepos)
+  program  = ["bash", "./get-last-commit.sh", each.key]
+}
+
+# update gitlab repo if github repo has new commits
+resource "null_resource" "update-repo" {
+  for_each = toset(local.filteredRepos)
+  triggers = {
+    last_commit = data.external.repos[each.key].result.last_commit
+  }
+  provisioner "local-exec" {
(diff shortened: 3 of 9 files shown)


terraform/github-to-gitlab-mirror/update-repo.sh
```diff
@@ -0,0 +1,17 @@
...
+    exit 1
+fi
+
+REPO_NAME=$1
+tempdir=$(mktemp -d)
+trap "rm -rf '$tempdir'" EXIT
+cd "$tempdir"
+
+git clone --mirror "https://github.com/$REPO_NAME" .
+git push --force --mirror "https://gitlab.com/$REPO_NAME"
```
Mic92, line 17: Not yet quite sure how this can be done in CI.
We would probably need a bot account.

terraform/github-to-gitlab-mirror/main.tf
```diff
@@ -0,0 +1,56 @@
...
+resource "gitlab_project" "repos" {
+  for_each               = toset(local.filteredRepos)
+  name                   = element(split("/", each.key), 1)
+  namespace_id           = gitlab_group.group.id
+  import_url             = "https://github.com/${each.key}"
+  shared_runners_enabled = false
+  visibility_level       = "public"
+}
+
+# when the api, we run into api rate limits
```
zowoq, line 41: Does this mean we cann't use webhooks for this?

Mic92: Clarified the gibberish I wrote yesterday. Pull mirroring is not available to free gitlab accounts unless you enabled the setting before they made it a premium feature.

terraform/github-to-gitlab-mirror/update-repo.sh
zowoq: Could use a hercules effect on a schedule?

Mic92: Ah nice. Haven't seen this feature: https://blog.hercules-ci.com/2022/11/15/easy-updates-with-onschedule-0.9.8/

zimbatm (approved):
very cool

bors merge

zowoq:
bors cancel

zowoq:
I hadn't reviewed this yet as I was expecting that the PR would be updated so it ran in CI before it was merged.

If we are going to merge it as is without CI I will make a few comments on it first.

terraform/github-to-gitlab-mirror/.envrc.private-template
```diff
@@ -0,0 +1,2 @@
+# Got to https://gitlab.com/-/profile/personal_access_tokens
+export GITLAB_TOKEN=<your_token>
```
zowoq, line 2: We could add https://github.com/nix-infra-bot as a gitlab admin and put the token in sops?

terraform/github-to-gitlab-mirror/main.tf
```diff
@@ -0,0 +1,59 @@
...
+  query = "user:${local.user} is:public archived:false"
+}
+
+resource "gitlab_group" "group" {
+  name             = local.user
+  path             = local.user
+  visibility_level = "public"
+}
+
+resource "gitlab_project" "repos" {
```
zowoq, line 32: Can we set `issues_enabled = false` and `merge_requests_enabled = false` on these repos?

terraform/github-to-gitlab-mirror/.envrc.private-template
Mic92: Well the equivalent for gitlab.com, so a new account.

terraform/github-to-gitlab-mirror/main.tf
```diff
@@ -0,0 +1,59 @@
...
+    "poetry2nix",
+  ]
+  # remove blacklist
+  filteredRepos = [
+    for repo in data.github_repositories.repos.full_names : repo if !contains(local.blacklist, element(split("/", repo), 1))
+  ]
+}
+
+data "github_repositories" "repos" {
+  query = "user:${local.user} is:public"
```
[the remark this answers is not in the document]
Mic92: Not really a concern in this repo but in general you cannot use the import function for private repos as far as I remember (or maybe if you provide credentials)

terraform/github-to-gitlab-mirror/.envrc.private-template
zowoq: > Well the equivalent for gitlab.com, so a new account.

Sorry, I don't understand?

Mic92: Ah. I didn't thought about the gitlab single-sign-on.

terraform/github-to-gitlab-mirror/get-last-commit.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+if [ -z "$1" ]; then
+  echo "Usage: $0 github_repo"
+  exit 1
+fi
+github_repo=$1
+last_commit=$(git ls-remote "https://github.com/$github_repo" HEAD | awk '{ print $1}')
```
zowoq, line 8: Not sure how we could catch new commits that aren't on the default branch?

How big a job would it be if we dropped this script and just always pushed the repos?

Mic92:
Don't have the energy for this just now. Anyone can take this over if they want.

zowoq:
@Mic92  Not sure if I'm misunderstanding it, could you please clarify how it works?

It's using github as a data source and stores state in tf cloud, wouldn't deleting a repo on github mean that it's missing in the next terraform run and the corresponding `gitlab_project` resource is destroyed?

Mic92:
Yes. That would be the case.

zowoq:
To run this on a schedule we'd need it to not destroy the `gitlab_project`, not sure of the best way to do that.

zowoq:
@Mic92 Do you mind if we don't use terraform for this? I'm thinking just writing a script for it will be easier.

PR #522, opened 2023-04-16
drop bors, switch to merge queue 

zowoq:


.github/workflows/flake-updates.yml
```diff
@@ -15,6 +15,11 @@ jobs:
           extra_nix_config: |
             experimental-features = nix-command flakes
       - name: Update flake.lock
+        id: update
         uses: DeterminateSystems/update-flake-lock@v19
         with:
           token: ${{ secrets.GH_TOKEN_FOR_UPDATES }}
+      - name: Enable Automerge
+        run: gh pr merge --rebase --auto "${{ steps.update.outputs.pull-request-number }}"
```
Mic92, line 23: Probably not a concern for this repository but how does merge queues work with pull requests from non-committers? Bors performs a snapshot of the code before it merges while merge queue seems to allow changes after being enabled.

zowoq: > merge queue seems to allow changes after being enabled.

I haven't actually used the merge queue yet so I'll have to check this, I had assumed that it wouldn't allow it.

zowoq:
https://github.com/orgs/community/discussions/46757

Reading though some of the feedback it seems this may still be a bit too new and buggy for us to use at the moment.

zowoq:
> https://bors.tech/newsletter/2023/05/01/tmib-76/
> As the primary maintainer, I officially declare Bors-NG to be feature frozen and deprecated. The public instance will remain running for awhile, but it will eventually start emitting warnings and be taken down. The source code will always be available if you want to run your own instance.

Mic92:
I guess I would again need something like this https://github.com/autifyhq/merge-queue-action to trigger automatic dependency merges.

zowoq:
> I guess I would again need something like this https://github.com/autifyhq/merge-queue-action to trigger automatic dependency merges.

Could use mergify for this? Could probably use it for this repo as well instead of the `gh pr merge` with the token?

zimbatm:
What do you think about deploying buildbot instead? @Mic92 has a working solution both for CI and also auto-merges, plus it's open-source and in a language that we can work with.

Mic92:
I think the only hurdle to solve is allowing users to add/remove their repository: https://hercules-ci.com/github/nix-community

zimbatm:
We could start with one repo, like this one, and then take it from there. It's just a shame for all the effort @zowoq put into converting the repo to Hercules CI, but I think it will be less effort long term.

Mic92:
Maybe I can reuse his work by also adding support for Hercules effects.

zowoq:
> What do you think about deploying buildbot instead?

I don't understand why you are proposing switching to buildbot?

zowoq:
https://blog.hercules-ci.com/2022/11/15/easy-updates-with-onschedule-0.9.8/

Hercules can create and merge flake update PRs by itself without a token, reason I wasn't going it use it here (yet) is I wanted to keep both flake update methods the same. 

I was planning to use this  in the docker and images repo so we don't need to deal with managing github tokens.

.github/workflows/flake-updates.yml
zowoq: I think if the PR is modified it's removed from the queue?

zowoq:
May as well merge this now so we can trial it before bors does actually need to be replaced.

roles/hercules-ci/secrets.yaml
```diff
@@ -1,6 +1,6 @@
...
-hercules-secrets: ENC[AES256_GCM,data:CG4W3mfGHS1bS3aNKRJHWWUmT9F5KB3hCYWSpPZDSElb6J+jhiIKP+d9n81yZ1T1cH4LpkULR8rsu00ZULMdRfghCF+KG3dperxYsMahhU0CjUj5mhCKb+6UpGOct87qJ3ILydFtBnVfkv4ohAjOdlidNDRnf4Yc0XKz1srBtBr0pmPf9hKMPXScs3nR4ecrkZjhC61CBCAVEuoGD+C8xklgV+v1Bg3CqjK2Rhxyhg5PfNZNtue5V7Lhq9TrY0S/9N+Hw3YxjAo4qCBQmOdLcLyN7MYwTJUmlIrPDzFAyfbXLRZZR6xX7+ArlgUbGNMlV3FjLXZ2E2+Abxp/uJ2/QF/6eKc44jp82hF7wjjc1My3Z//sT3pZZas/dVe40n715CtkzN8bdBZw/fHxc61QhDMz4wBpQ+SPZLdiHZ67COAkw/XKBdym4GEziqWEuYrx1aEgj5k4+qXOFgjqKHXeT783ZK2YDTwtJieSy/2bQCsTYg2V1GUDPEjxPqptDSDVDlTbMJ05EHELQvKpWQ8/cuQonGHdA9ENldA43rHnL98VLuS+YLToM0LN2oUGfXElISTXtt+x2bfBvOrNwC28nVQTCo1pgQMMWI3Tn1u2VDETBq3/X5hgOh+wDvikFnQEmw+DDd/b7eA3B4i0u/fS33aHprqqsjKdaYPmeFtQ4KIY8ZCaj8CcReH8Xp48E4JklVS2/BaPRnv7TlHQAhYKD+KMlRub9pzhhwNpRlkum/v+q1g0ZVr7kVB8e7Y=,iv:+dILru8YNvl51tFGoOTtYwpaTbcuiujtkX4zTAzqDlg=,tag:N6VAPhSha/6cjZW/EvYKcw==,type:str]
```
zowoq, line 3: Removed the branch condition, doesn't look like hercules can match `refs/heads/gh-readonly-queue/master/*`
```
{ "or": [{ "isBranch": "staging" }, { "isBranch": "trying" }] }
```

zowoq:
I plan to merge this in 48hrs.

PR #537, opened 2023-04-21
roles/common: various

zowoq:


roles/common.nix
```diff
@@ -7,11 +7,6 @@
     ./users.nix
   ];
 
-  # Nicer interactive shell
-  programs.fish.enable = true;
-  # And for the zsh peeps
-  programs.zsh.enable = true;
-
```
Mic92, line 14: Do we have this still for build01?

zowoq: Do we need other shells? e.g. the aarch64 build box doesn't have them.

Fish seems to be building a bunch of completion stuff as separate drvs which seems a bit pointless on these machines.

roles/common.nix
```diff
@@ -22,7 +17,14 @@
   systemd.services.mdmonitor.enable = false;
 
   # speed-up evaluation & save disk space by disabling manpages
-  documentation.enable = false;
+  documentation = {
+    enable = false;
+    man.enable = false;
+    info.enable = false;
+    nixos.enable = false;
+  };
```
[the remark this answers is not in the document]
zowoq: > is `documentation.enable = false` not inferring all of the rest?

IIRC it did but noticed that mandb cache was still getting built for so I'm disabling all of them so they can't leak back in.

> Let's push that to srvos if not.

To me this doesn't read like a request but instead an instruction.

I'd be fine with how it is phrased if we were talking about a nix-community project but srvos is not a nix-community project.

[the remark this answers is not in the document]
zimbatm: Ok. My intention was to make this a proposal or a suggestion. Thanks for the PR by the way.

roles/common.nix
adisbladis: > Fish seems to be building a bunch of completion stuff as separate drvs which seems a bit pointless.

It's because it can generate completions from man pages.

zowoq: Moved fish/zsh to roles/buillder for build01 and disabled the generated completion.

roles/common.nix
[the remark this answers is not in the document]
zowoq: Dropped the documentation options from this PR so we're using the srvos settings.

PR #540, opened 2023-04-25
import ssh_host_ed25519_key

zowoq:


tasks.py
```diff
@@ -184,6 +168,63 @@ def deploy(c, hosts=""):
...
+    Decrypt host private key, install with nixos-anywhere. Use inv install --hosts build01
+    """
+    g = DeployGroup(get_hosts(hosts))
+
+    def anywhere(h: DeployHost) -> None:
+        hostname = h.host.replace(".nix-community.org", "")
+        with tempfile.TemporaryDirectory() as tmpdir:
+            decrypt_host_keys(c, hostname, tmpdir)
+            c.run(
+                f"nix run github:numtide/nixos-anywhere#nixos-anywhere -- --extra-files {tmpdir} --flake .#{hostname} {h.host}"
```
zowoq, line 222: This won't work with aarch64-linux as the default kexec tarball is x86_64.

@Mic92 Could nixos-anywhere be changed to read the system from the flake and set the right tarball?

zowoq:
Moved the tasks.py changes to the disko PR so this just imports the current ssh_host_ed25519_key and prevents rsa keys from being created.

zimbatm (approved):

tasks.py
Mic92: Sounds like a good idea until $someone finalizes aarch64 ci.

PR #542, opened 2023-04-25
switch to mkdocs 

zowoq:
The support sections are already out of sync as I forgot to add the admin email to docs/README.

Easier if we only have have support section to update.


zimbatm:
What do you think of pushing everything in the ./docs instead? That way website visitors have access to all the info without being redirected. Or we redirect the whole website directly to the README?

zowoq:
> What do you think of pushing everything in the ./docs instead?

I had been thinking of doing this and also trying to make the website a bit nicer looking. Not sure if we'd want to stay with jekyll or switch to something else? Also some method of link checking would be good.

zowoq:
I was playing around with mkdocs and ended up redoing the website.

Would be deployed to `gh-pages` by hercules when PRs are merged to master.

Need to update the settings in https://github.com/nix-community/infra/settings/pages if we merge this.

zimbatm:
yup, this is moving in the right direction. it makes room for even more documentation of all of the practices that we have.

zowoq:
This is tidied up enough that it's mergable, links should be working and formatting adjusted so it doesn't look incorrect.

zimbatm (approved):
:+1:

PR #544, opened 2023-04-27
terraform: manage github org admins

zowoq:
This'll make onboarding much simpler.

For managing non-admin members, repos, teams, etc I expect that we'd want to have that in a separate workspace or a separate repo.

I was going to do gitlab here as well but I'll do it in a separate PR.

### DIFF
--- a/terraform/locals.tf
+++ b/terraform/locals.tf
@@ -7,4 +7,8 @@ locals {
     zimbatm    = "zimbatm@zimbatm.com"
     zowoq      = "zowoq.gh@gmail.com"
   }
+
+  bot = {
+    nix-infra-bot = "admin@nix-community.org"
+  }
 }
--- a/terraform/flake.nix
+++ b/terraform/flake.nix
@@ -12,6 +12,7 @@
               p.cloudflare
               p.external
               p.gandi
+              p.github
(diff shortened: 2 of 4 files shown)


terraform/locals.tf
```diff
@@ -7,4 +7,8 @@ locals {
     zimbatm    = "zimbatm@zimbatm.com"
     zowoq      = "zowoq.gh@gmail.com"
   }
+
+  bot = {
+    nix-infra-bot = "admin@nix-community.org"
```
zowoq, line 12: @Mic92 Can we transfer this account so it is controlled by the org and we store the login in sops? Could create another github account if that is easier?

Mic92: done.

zowoq: Thanks, I'll reset the password, add 2fa and add it to secrets.yaml in the repo root.

zowoq:
@zimbatm We decided against using terraform to manage the github org admins in https://github.com/nix-community/infra/pull/563#issuecomment-1537357542, what about the gitlab org admins? The security concerns probably aren't as important there? I'd like to reduce the number of manual onboarding tasks.

Issue #541, opened 2023-04-25
nixpkgs-update logs backup


zowoq:
https://r.ryantm.com/log/

I was going to do a manual backup of these logs before reinstalling build02 but I suppose we should really be doing regular backups so they aren't all lost if build02 dies.

If we don't have somewhere external to nix-community to back them up to I guess we'd want backups on a couple of our machines?

cc @nix-community/admin

Mic92:
We might have backup storage provided by hetzner that comes with the hardware. Their storage boxes support borgbackup.

zowoq:
> We might have backup storage provided by hetzner that comes with the hardware.

If we do have this it doesn't seem to be accessible with the credentials in the top level secrets so would need someone else with access to deal with it on the hetnzer side.

> Their storage boxes support borgbackup.

This looks like it'll do what we want and it's simple to set up, hopefully we can use it.

Mic92:
I'll have a look at it.

zowoq:
I'll set this up in the next couple of days.

PR #550, opened 2023-04-30
build02: backup nixpkgs-update logs

zowoq:
Closes https://github.com/nix-community/infra/issues/541

@Mic92 I copied most of this from your dotfiles as I noticed that you used to have a borgbackup storagebox.

cc @nix-community/admin

### DIFF
--- a/build02/nixpkgs-update-backup.nix
+++ b/build02/nixpkgs-update-backup.nix
@@ -1,0 +1,46 @@
+{ config, ... }:
+{
+  # 100GB storagebox is under the build03 hetzner account
+
+  sops.secrets.hetzner-build03-borgbackup-ssh = { };
+
+  programs.ssh.knownHosts = {
+    "hetzner-storage-box" = {
+      hostNames = [ "[u348918.your-storagebox.de]:23" ];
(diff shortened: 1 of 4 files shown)


build02/nixpkgs-update-backup.nix
```diff
@@ -0,0 +1,46 @@
...
+  systemd.services.borgbackup-job-nixpkgs-update.serviceConfig.ReadWritePaths = [
+    "/var/log/telegraf"
+  ];
+
+  services.borgbackup.jobs.nixpkgs-update = {
+    paths = [
+      "/var/log/nixpkgs-update"
+    ];
+    repo = "u348918@u348918.your-storagebox.de:/./nixpkgs-update";
+    encryption.mode = "none";
```
zowoq, line 23: I don't think we need to bother with encryption?

Mic92:
Do you have full access to this storage box or do you need further credentials / setup from my side?

build02/nixpkgs-update-backup.nix
Mic92: Probably yes since the logs are public.

zowoq:
Shouldn't require anything else, I was able to set a password and copy the ssh key.

PR #559, opened 2023-05-05
modules/nixos: add comin

zowoq:
Polling the git repo looks much nicer than `autoUpgrade` that we're currently using.

I'd like to trial it on build01 but need to address a couple of thing first.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -26,6 +26,26 @@
         "type": "github"
       }
     },
+    "comin": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1710799524,
+        "narHash": "sha256-5ZMHWRYsj3ybp2J+tsSpiMIcotI7CHngAx6NdLfSPl8=",
+        "owner": "nlewo",
+        "repo": "comin",
+        "rev": "b4f7dcc7c0f1a6ad6470c762f1ddccfb7ee887e4",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nlewo",
+        "repo": "comin",
+        "type": "github"
+      }
+    },
     "disko": {
       "inputs": {
         "nixpkgs": [
@@ -179,6 +199,7 @@
     "root": {
       "inputs": {
         "buildbot-nix": "buildbot-nix",
+        "comin": "comin",
         "disko": "disko",
         "flake-parts": "flake-parts",
         "nix-darwin": "nix-darwin",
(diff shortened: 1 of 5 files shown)


nlewo:
That could still be a bit early since i didn't do any release yet. So i was still feeling free to break interfaces ;)

But do not hesitate to create issues with features you would like to see.

Mic92:
How does it make deployment failures visible? Does it make the service fail? I also made the experience that many service upgrades failures can be mitigated by just adding one retry of nixos-rebuild. So it would be cool if it does that.

nlewo:
> How does it make deployment failures visible? 

I'm planning to provide a Prometheus metrics. Currently, comin reports its status over http and when the `head_commit_deployed` attribute is `false`, it means an issue occurred during the deployment.

> Does it make the service fail? 

No (it is a daemon).

> I also made the experience that many service upgrades failures can be mitigated by just adding one retry of nixos-rebuild. So it would be cool if it does that.

Noted (that should not be hard to add this option). 
I also had planned to address this kind of needs by adding the `force=true` parameter to the http route `/deploy` (currently used to provide by webhooks).

zowoq:
We talked about deployments a bit more recently and looks like we'll switch `cachix deploy` as they going to add telemetry and have support for nix-darwin.

Sorry for the noise here @nlewo.

nlewo:
@zowoq no worries and thank you for your explanations.
(Moreover, comin is definitively not ready for being used by others. Sometimes, i "forget" to publish my softwares... but for the first time, i published it too early :/)

zowoq:
The `cachix deploy` telemetry didn't end up working out so I'd like to try this now that it has had a release.

Would need to use http for the status but that doesn't look to be hard to add to our monitoring.

nlewo:
> time="2024-03-07T04:48:51Z" level=info msg="Evaluation error: The defined comin.machineId (2a7f74e3b4f6409eb031a1e606fff1bc) is different to the machine id () of this machine"

This is an issue in comin which should be fixed by https://github.com/nlewo/comin/pull/14.

> Would need to use http for the status but that doesn't look to be hard to add to our monitoring.

Which metrics would you like to collect?
In case you use Prometheus, i created https://github.com/nlewo/comin/issues/9 to collect a list of useful metrics. Maybe that could fit your need better than the unstable `status` output.

zowoq:
> This is an issue in comin which should be fixed by ...

Thank you!

> Which metrics would you like to collect?

I think a status/exit code of the most recent deployment and the commit of the last successful deployment would be enough for us. 

> In case you use Prometheus ...

Yes, we're using prometheus so metrics in that format would be excellent.

zowoq:
I'll test this on build01 for a bit and also so I have a live machine to work on while I figure out the the http status monitoring.

Issue #398, opened 2023-01-22
Configure Github with Terraform


zimbatm:
We have the [Settings app](https://github.com/apps/settings) for repository owners to self-configure their repos.

What's missing is to describe all of the teams and members as code to increase transparency even further.

That also helps create a membership list, potentially allow members to send PRs to add new repos, ...

zowoq:
I've already been looking at this, short term it seems good but I'm a bit concerned that it will end up being a burden to maintain long term.

At the moment I'm thinking about dumping the org api daily and pushing it to a  repo, which would give us a transparent log.

zimbatm:
We could start with partial coverage. One thing that would be good is to have this 1:1 mapping between teams and repo names.

zowoq:
https://github.com/ipdxco/github-as-code (from https://github.com/NixOS/org)

This looks like it could be good but:

1. s3 only, I don't think we'd want to start giving money to aws.
1. assuming it wasn't s3 only we still couldn't use it, terraform cloud free tier is limited to 500 resources.

May need to use gitlab for state or self host (IIRC all of the other free TACOS have some sort of resource/user limit), would also want to keep state locking.

zowoq:
terraform 1.11 (current release) and opentofu 1.10 (next release) both support s3 native locking so moving forward with `github-as-code` may be feasible if we self host s3.

PR #563, opened 2023-05-05
secrets.yaml: add nix-infra-bot

zowoq:
After this is merged I'll make `nix-infra-bot` an org owner and generate an token with org permissions to use with terraform in https://github.com/nix-community/infra/pull/544.

I'll do the same again in another PR for gitlab.


zimbatm:
That seems pretty dangerous. Making the bot an org owner weakens the security of the setup. For less frequent activities such as changing the list of admins, we can have the bot check, but should probably apply things manually.

zowoq:
> Making the bot an org owner weakens the security of the setup. For less frequent activities such as changing the list of admins, we can have the bot check, but should probably apply things manually.

How is https://github.com/nix-community/infra/issues/398 supposed to work?

zimbatm:
One approach could be to map all of the resources with Terraform still. For public resources, the secrets.GITHUB_TOKEN should be able to query them. In that scenario the CI can do the plan, and would fail on apply. When the failure happens, it's up to the admin to step in and apply the changes manually. It makes things slightly less convenient, but it less convenient, but it would also add an extra layer of security.

zowoq:
> When the failure happens, it's up to the admin to step in and apply the changes manually.

I'd rather manage the org manually and just dump the org api daily as a log.

zimbatm:
I wonder if we can forward the security audit to Matrix

zowoq:
iirc streaming the audit log is a paid feature.

We could try to replicate it by dumping the api hourly, pushing it to a repo and posting the diff?

Mic92:
Github's audit log could be used as well:

https://github.com/organizations/nix-community/settings/audit-log?q=action%3Aorg.invite_member

zowoq:
> Github's audit log could be used as well:
> 
> https://github.com/organizations/nix-community/settings/audit-log?q=action%3Aorg.invite_member

That's what I was referring to. We can look at it and export it manually but I think api access to it is a paid feature.

Mic92:
Might be overkill for now I think to extract it automatically. If something wrong happens we could still visit the page.

PR #564, opened 2023-05-07
modules/nixos: github-org-backup 

zowoq:
Now that we have access to external storage we could use it to backup the repos.

https://github.com/gabrie30/ghorg

~~Still need to add secrets, I'll do that if we want to move forward with this.~~

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -156,6 +156,7 @@
 
           cachix-deploy = ./modules/nixos/cachix-deploy;
           community-builder = ./modules/nixos/community-builder;
+          github-org-backup = ./modules/nixos/github-org-backup.nix;
           hercules-ci = ./modules/nixos/hercules-ci;
           hydra = ./modules/nixos/hydra.nix;
           nur-update = ./modules/nixos/nur-update.nix;
--- a/hosts/build03/configuration.nix
+++ b/hosts/build03/configuration.nix
@@ -21,6 +21,7 @@
     inputs.self.nixosModules.remote-builder-darwin02
     inputs.self.nixosModules.remote-builder-darwin03
 
+    inputs.self.nixosModules.github-org-backup
     inputs.self.nixosModules.hydra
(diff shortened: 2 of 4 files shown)


services/github-backup/default.nix
```diff
@@ -0,0 +1,59 @@
...
+  systemd.services.github-backup = {
+    script = ''
+      GHORG_GITHUB_TOKEN="$(<$CREDENTIALS_DIRECTORY/github-backup-token)" \
+        ${pkgs.lib.getExe pkgs.ghorg} clone nix-community \
+        --backup \
+        --clone-wiki \
+        --concurrency 2 \
+        --exclude-match-regex '^nixpkgs$' \
+        --exclude-match-regex '^nix$' \
+        --path /var/lib/github-backup \
```
Mic92, line 14: Since this is public information, is there a way to make this backup public accessible?

zowoq: Do you mean serving this folder with nginx or something else?

Mic92: Something like that. I just don't know how easy it is to download the backup over nginx.

zowoq: Can you explain why you want the backups to be publicly accessible? My understanding was running a mirror wasn't the objective, it was just a way of being able to have backups?

zowoq:
I'll finish

services/github-backup/default.nix
adisbladis: We could expose the storage box from `web01` if we want it publicly accessible.

Mic92: with its own subdomain.

zowoq: > Can you explain why you want the backups to be publicly accessible? My understanding was running a mirror wasn't the objective, it was just a way of being able to have backups?

I'd still like an answer to this?

PR #570, opened 2023-05-09
docs: separate open collective and sponsors

zowoq:
Seems the open collective svg stopped working as they aren't showing the backers and sponsors themselves:
collective1

___

Switched to the js banner and moved the collective to a separate page to give it more space:

collective2


zimbatm (approved):
thanks!

PR #584, opened 2023-05-15
flake: split up per host

zimbatm:
Inject the inputs in specialArgs so they can be used in the modules directly. This allows us to split up the top-level flake as well.

### DIFF
--- a/build03/default.nix
+++ b/build03/default.nix
@@ -1,0 +1,7 @@
+{ self, ... }:
+{
+  flake.nixosConfiguration.build03 = self.lib.nixosSystem {
+    system = "x86_64-linux";
+    modules = [ ./configuration.nix ];
+  };
+}
--- a/flake.nix
+++ b/flake.nix
@@ -42,107 +42,64 @@
   };
 
   outputs = inputs @ { flake-parts, self, ... }:
-    flake-parts.lib.mkFlake
-      { inherit inputs; }
-      {
-        systems = [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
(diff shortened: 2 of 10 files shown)


Mic92 (approved):

zowoq:
Adding default.nix files with a couple of lines is not a good pattern.

Very strong 👎, I want to revert this.

flake.nix
```diff
@@ -42,107 +42,64 @@
...
-                  ./build04/configuration.nix
-                  inputs.disko.nixosModules.disko
-                ];
-            };
-          };
+      flake.lib.nixosSystem = args:
+        inputs.nixpkgs.lib.nixosSystem ({ specialArgs = { inherit inputs; }; } // args);
+
+      flake.nixosModules = {
+        common = ./roles/common.nix;
```
zowoq, line 102: What's this for?

zowoq:
I'm merging the revert as `nixosConfigurations.build03` is missing since this PR.

build03/default.nix
```diff
@@ -0,0 +1,7 @@
+{ self, ... }:
+{
+  flake.nixosConfiguration.build03 = self.lib.nixosSystem {
```
zowoq, line 3: `nixosConfiguration` -> `nixosConfigurations`

I would have expected this to error rather than be ignored.

Maybe we need to switch back to the ci.nix style list of attributes we were using previously to avoid this.

flake.nix
zimbatm: Hmm, this didn't belong in this PR. It's related to the deeper structure changes I started working on. See  https://github.com/nix-community/infra/pull/589

PR #589, opened 2023-05-16
introduce mac01

zimbatm:
Move the code around so it supports both NixOS and nix-darwin. This is a
pretty big move that probably will require some cleanup afterwards.


zowoq:
We're using darwin01.nix-community.org for https://github.com/winterqt/darwin-build-box.

zimbatm:
ok, we probably want a naming scheme to distinguish between community boxes vs CI boxes. Now that we have multiple architectures, `build01` isn't descriptive enough anymore.

zowoq:
> This is a pretty big move that probably will require some cleanup afterwards.

I'd really prefer to slow down a bit and not do this all in one big PR.


> distinguish between community boxes vs CI boxes

Is keeping these separate still necessary after we have a nix release with "Trustless remote building" PR?

zimbatm:
Generally, I use Draft PRs to get high-level feedback.

For now, I mostly care about feedback on splitting the modules under `modules/{shared,darwin,nixos}`. Where "shared" is compatible with both module systems". The goal is to make room for mac01, or however it will be called. Once that's ready I can split it out in another PR.

Another sub-goal is to reduce the number of times `../` is used in the Nix code. In my experience, refactoring gets easier when absolute references are being used, as they are easier to grep around.

zimbatm:
> Is keeping these separate still necessary after we have a nix release with "Trustless remote building" PR? Don't remember if it helps with that or if it just means we can be less strict with who gets access to the builder?

The security posture is not super clear to me. Darwin builds are not sandboxed so it might make sense to be more careful about sharing those.

zowoq:
I'd like to add the new hardware first then once it's stable we can dedupe and refactor everything. I don't want to do both at the same time.

zowoq:
> Another sub-goal is to reduce the number of times ../ is used in the Nix code. In my experience, refactoring gets easier when absolute references are being used, as they are easier to grep around.

Would be easier to this first for the existing code and would reduce the diff of this PR a lot.

zowoq:
I think we've implemented everything we needed from this PR, closing.

PR #592, opened 2023-05-16
darwin02: hercules with remote builder

zowoq:


zowoq:
`nix-darwin` added `openssh.authorizedKeys.keys` recently so we don't need to manage them manually.

zimbatm:
@zowoq is anything missing on this PR? I think @DavHau's project is blocked on this.

zowoq:
> is anything missing on this PR?

Yes, ssh keys, builder config, etc. I basically stopped working on this as you were busy, I've been waiting for you come back and sort out access to the machine.

zimbatm:
Ok, do you mind taking over? You should be able to `ssh m1@51.159.120.155`

zowoq:
> Ok, do you mind taking over?

Sure, I should be able to get it finished in the next few days.

> You should be able to `ssh m1@51.159.120.155`

It's not working with my ssh key if that's what you mean? `m1@51.159.120.155: Permission denied (publickey)`

I'll see how far I can get but iirc some of the mac settings need approval via GUI so may need access to that or someone to do it for me.

zowoq:
@roberth Could you help with this please? 

We were trying to avoid having secrets on the darwin machine and use it just as a remote builder.

https://docs.hercules-ci.com/hercules-ci/guides/remote-builder#_configure_nix_to_use_the_remote_builder
```
sudo -u hercules-ci-agent nix-build --expr 'with import <nixpkgs> { system = "aarch64-darwin"; }; hello.overrideAttrs (o: { n = 123; })'
```

`darwin02` so is configured as a remote for `build03`, the above command works on `build03` but still getting `No agent available` on the dashboard.

> While this protocol isn’t as flexible as the agent’s protocol, it’s still useful for builds in Effects. If you need to build derivations in an effect, and those derivations' system does not match that of the effect, you currently need to set up a remote builder.

I overlooked this bit initially, does this mean the remote builder is only for effects and doesn't work for normal builds? So we would need to install the agent on the darwin machine?

zowoq:
Might revisit this but if we need secrets on the darwin hosts for cachix deploy not much point in trying to avoid them for this.

PR #593, opened 2023-05-16
roles, services: refactor into modules

zowoq:


build04/configuration.nix
```diff
@@ -3,15 +3,15 @@
   imports = [
     inputs.disko.nixosModules.disko
     ./hardware-configuration.nix
-    ../roles/common.nix
-    ../roles/hercules-ci
-    ../roles/remote-builder/user.nix
+    inputs.self.modules.nixos.common
```
Mic92, line 6: Might be worth it to also pass self as a specialArg and use that directly.

flake.nix
```diff
@@ -112,6 +112,23 @@
             };
           };
 
+        flake.modules = {
```
Mic92, line 115: We have a standard attribute for that: https://nixos.wiki/wiki/Flakes#Output_schema

```suggestion
        flake.nixosModules = {
```

zowoq:
Moved the first commit to another PR: https://github.com/nix-community/infra/pull/605

build04/configuration.nix
zowoq: I'll leave this for now as we need inputs in a few places anyway.

zowoq:
I'll merge this in 24 hours.

zimbatm (approved):
LGTM

PR #594, opened 2023-05-16
effect, shell, treefmt: move into /dev

zowoq:
keeps the repo root tidy

I guess `effect.nix` isn't exactly "dev" but not sure where else to put it.


zimbatm (approved):

zowoq:
I'll leave the effect in the repo root for now and see if there is a nicer way of handling the paths.

PR #596, opened 2023-05-18
.github/settings.yml: make dpulls a required check

zowoq:
I'd like to enable https://github.com/marketplace/dpulls for this repo and make it a required check, it is useful for stacking PRs.

It's been used in https://github.com/NixOS/nix for while now.


zimbatm (approved):
Sure, why not

zowoq:
This probably isn't that useful, might only use it a couple of times. Sorry for the noise.

PR #597, opened 2023-05-18
build0*: move into /hosts

zowoq:
e.g.
```
hosts
  darwin02
  build01
  build02
  build03
  build04
```


zimbatm:
I don't get what problem it resolves since the machine name already gives a hint of what type of system it is.

zowoq:
Forgot to say why, sorry. Would like to push the machine dirs down a level instead of keeping them all in the repo root.

Both darwin and nixos in one directory could work as well?
```
system
  darwin02
  build01
  build02
  build03
  build04
```

zimbatm (approved):

zowoq:
No, it doesn't solve a technical problem, just easier to work with as we add more hardware.

PR #599, opened 2023-05-19
dev/shell: flatten

zowoq:
Same as we did for treefmt in https://github.com/nix-community/infra/pull/518.

### DIFF
--- a/dev/shell.nix
+++ b/dev/shell.nix
@@ -1,16 +1,15 @@
+{ pkgs, ... }:
 {
-  perSystem = { pkgs, ... }: {
-    devShells = {
-      default = with pkgs; mkShellNoCC {
-        packages = [
-          jq
-          python3.pkgs.deploykit
-          python3.pkgs.invoke
-          rsync
-          sops
-          ssh-to-age
-        ];
-      };
+  devShells = {
+    default = with pkgs; mkShellNoCC {
+      packages = [
+        jq
+        python3.pkgs.deploykit
+        python3.pkgs.invoke
+        rsync
+        sops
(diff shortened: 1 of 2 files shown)


dev/shell.nix
```diff
@@ -1,16 +1,15 @@
+{ pkgs, ... }:
```
zimbatm, line 1: A general observation: as a reader I don't know what the interface is until I see in what context that file is being consumed.

This is a common issue already between nixpkgs callPackage and NixOS modules. But this is making things worse.

I don't have a really good answer except maybe that each file might have to be annotated on the top with `# flake-parts perSystem module` or some qualifier like that.

zowoq: I don't see how this is a problem? The top level flake.nix provides context, don't see any benefit in adding comments to files?

zimbatm: It's sub-optimal because the reader doesn't have all the context when opening the file. In order to find out, they have to now search where the file is being used. It also makes the life of editors harder to provide auto-complete.

As we become more experienced with Nix, we learn to recognize those patterns and learn to infer that a file starting with `{ stdenv, ... }:`, for example, is probably loaded using `callPackage`.

But it's just a general observation and a particularity of Nix, not something we necessarily need to act upon.

zowoq: I'm really struggling to understand your point of view.

I'd say https://github.com/nix-community/infra/pull/518 and `build0*/default.nix` parts of https://github.com/nix-community/infra/pull/584 also removed context?

Honestly my take away from you comments is flattening files and splitting up the flake is bad?

PR #602, opened 2023-05-19
terraform: move nix-community workspace into /nix-community

zowoq:
Not needed at the moment but this would be the first step in adding another terraform workspace that shares dependencies an secrets.


Mic92:
How do you want to manage secrets?

zowoq:
Might just have one set of secrets that is shared between workspaces? 

Could try splitting them up but would need to either duplicate secrets or have another set of shared secrets.

```
/terraform/admin/secrets.yaml <- cloudflare
/terraform/nix-community/secrets.yaml <- cloudflare
```

```
/terraform/secrets.yaml <- cloudflare
/terraform/admin/secrets.yaml
/terraform/nix-community/secrets.yaml
```

zowoq:
Closing this as we aren't likely to actually need it for a while (if ever) and I don't want to increase depend on workspaces just in case we want to use a different backend that doesn't support workspaces, e.g. gitlab. (could maybe use separate state files there instead if we do end up wanting to split this up)

PR #617, opened 2023-06-04
.github/labeler.yml: add secret file in repo root

zowoq:


zowoq:
Seems the effect is stuck for some reason.

PR #619, opened 2023-06-05
darwin02: init

zowoq:


hosts/darwin02/builder.nix
```diff
@@ -0,0 +1,25 @@
...
+  users.users.nix = {
+    name = "nix";
+    uid = 8765;
+    home = "/Users/nix";
+    createHome = true;
+    shell = "/bin/zsh";
+    description = "User for remote build clients";
+    # keys are copied, not symlinked
+    openssh.authorizedKeys.keys = [
+      "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEmdo1x1QkRepZf7nSe+OdEWX+wOjkBLF70vX9F+xf68 builder"
```
zimbatm, line 20: It's possible to further restrict the access so the ssh user can only be used for remote building. See: 
https://github.com/numtide/srvos/blob/4f22e6fcaf17c6313c2ecdc996760c3e4b14a623/nixos/roles/nix-remote-builder.nix#LL32C8-L32C8

hosts/darwin02/builder.nix
```diff
@@ -0,0 +1,25 @@
...
+  users.knownGroups = [ "nix" ];
+  users.knownUsers = [ "nix" ];
+
+  users.groups.nix = {
+    name = "nix";
+    gid = 8765;
+    description = "Group for remote build clients";
+  };
+
+  users.users.nix = {
```
zimbatm, line 11: Now that it's deployed I don't know if it's worth changing it, but in srvos we named that user "nix-remote-builder" to clarify the role of that user. `nix` might be mistaken as the user for the nix-daemon.

zimbatm (approved):
nice. Since it's already deployed I propose to merge it directly and do potential follow-up cleanups down the road.

hosts/darwin02/builder.nix
zowoq: If we add darwin support to `srvos` I'm planning to adjust that module support darwin and then using it here for both darwin and linux.

hosts/darwin02/builder.nix
zowoq: I used `nix` here for consistency with our existing builder config but this would also be addressed by using the srvos module.

PR #628, opened 2023-06-10
modules/remote-builder: refactor, rename

zowoq:


modules/nixos/remote-builder/build04.nix
```diff
@@ -9,7 +9,7 @@
       sshKey = config.sops.secrets.id_buildfarm.path;
       sshUser = "nix";
       system = "aarch64-linux";
-      supportedFeatures = [ "big-parallel" ]; # sync with build04/configuration.nix
+      supportedFeatures = inputs.self.nixosConfigurations.build04.config.nix.settings.system-features;
```
Mic92, line 12: What's the impact on evaluation speed of this?

zowoq: Not much but I guess we may as well avoid it where we can, reverted this change.

PR #635, opened 2023-06-18
import modules 

zowoq:
Just doing a clean import in this PR, in subsequent PRs I'll share this code between darwin/linux and expose it as `{darwin,nixos}Modules`.


zimbatm:
What would upstream need to have so it's not needed to fork the content?

zowoq:
@zimbatm There has been discussion about this in the community monitoring room, probably best if we continue there.

PR #644, opened 2023-06-27
terraform/cloudflare: remove search.nix-community.org

zowoq:
We removed hound seven weeks ago so I think anyone using it will have noticed the redirect by now.

Also github moved code search behind a login a couple of weeks ago so the redirect is less useful than it was originally.

I'll delete the redirect once the PR is merged.

### DIFF
--- a/terraform/cloudflare_nix-community_org.tf
+++ b/terraform/cloudflare_nix-community_org.tf
@@ -113,27 +113,6 @@ resource "cloudflare_record" "nix-community-org-nur-update-CNAME" {
   type    = "CNAME"
 }
 
-resource "cloudflare_record" "nix-community-org-search-CNAME" {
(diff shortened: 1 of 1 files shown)


zimbatm:
ok!

terraform/cloudflare_nix-community_org.tf
```diff
@@ -113,27 +113,6 @@ resource "cloudflare_record" "nix-community-org-nur-update-CNAME" {
...
-  zone_id = local.nix_community_zone_id
-  name    = "search"
-  value   = "nix-community.org"
-  type    = "CNAME"
-  proxied = true
-}
-
-# created manually
-#
-#resource "cloudflare_page_rule" "nix-community-org-search-forward" {
```
zowoq, line 126: This has been removed.

PR #647, opened 2023-06-29
.sops.yml: Add adisbladis key

adisbladis:


.sops.yaml
```diff
@@ -9,6 +9,7 @@ keys:
   - &ryantm age1d87z3zqlv6ullnzyng8l722xzxwqr677csacf3zf3l28dau7avfs6pc7ay
   - &zimbatm age1jrh8yyq3swjru09s75s4mspu0mphh7h6z54z946raa9wx3pcdegq0x8t4h
   - &zowoq age1m7xhem3qll35d539f364pm6txexvnp6k0tk34d8jxu4ry3pptv7smm0k5n
+  - &adisbladis 00244EF5295026AA323A4BDB110BFAD44C6249B7
```
zowoq, line 12: This isn't a valid key?

zowoq: Ah, gpg. We got rid of that, could you please use an age key instead?

adisbladis: I find using age keys more inconvenient as I don't already have that integrated into my key management workflow.
I've generated an age key but I think it's a pretty massive PITA to manage more keys.

adisbladis: To elaborate a bit on as to _why_ it is a PITA:
I have a strict policy not to have keys on disk but only in crypto hardware such as a Yubikey.

My GPG was generated in an air gapped environment and I have physical backups residing in other Yubikey's & one that can be recovered by assembling all the parts (https://en.wikipedia.org/wiki/Shamir%27s_secret_sharing) which are spread in multiple physical locations.
Setting up this kind of key management is not exactly trivial and takes a lot of effort.
Since I went through this effort in 2017 I have not had to do it again.

This key was now generated ad-hoc and I'm going to have to regenerate it again and give it the full treatment.

.sops.yaml
```diff
@@ -9,6 +9,7 @@ keys:
   - &ryantm age1d87z3zqlv6ullnzyng8l722xzxwqr677csacf3zf3l28dau7avfs6pc7ay
   - &zimbatm age1jrh8yyq3swjru09s75s4mspu0mphh7h6z54z946raa9wx3pcdegq0x8t4h
   - &zowoq age1m7xhem3qll35d539f364pm6txexvnp6k0tk34d8jxu4ry3pptv7smm0k5n
+  - &adisbladis age1yubikey1qt7xflk6z7sqp28tuyp0y239l9dklllklaxffqqln8dtvvurxp4m5hgx9nd
```
zowoq, line 12: Seems `sops` doesn't support this type of key.

```
failed to parse input as Bech32-encoded age public key: malformed recipient "age1yubikey1qt7xflk6z7sqp28tuyp0y239l9dklllklaxffqqln8dtvvurxp4m5hgx9nd": invalid type "age1yubikey"
```

I'd prefer to not re-add the gpg stuff, TLDR: for me it is just extra stuff to download every time I use the shell.nix.

That said if you can't use a regular age key you can see sops gpg hook that you'll need to add in https://github.com/nix-community/infra/commit/92a43cb33d88ad5983f34f26818a0f23e0ebf03c.

zowoq: Could you use your current ssh key instead of generating another key? https://github.com/Mic92/ssh-to-age

Mic92: Yes. Sops does not support yubikey keys

adisbladis: > Could you use your current ssh key instead of generating another key? https://github.com/Mic92/ssh-to-age

I do not use file based ssh keys. I use `gpg-agent` and my key resides in hardware.

adisbladis: So the status quo is that because GPG keys are unsupported here for some reason I fail to understand, the yubikey age plugin isn't supported by sops hardware based keys are not supported.

zimbatm: Would it be ok to generate an age key specifically to deploy to the nix-community? It doesn't strictly need to be correlated with your SSH key AFAIK.

Most of us aren't using GPG so we saw an opportunity to remove it from our system config. I think you were AWOL at the time we were discussing this so we missed your input.

Mic92: I would be OK adding gnupg support back.

zimbatm: Same, I still have GPG installed due to gopass

zowoq: Opposed to re-adding it just for 1/5 users and 1/11 total keys.

adisbladis: > Would it be ok to generate an age key specifically to deploy to the nix-community? It doesn't strictly need to be correlated with your SSH key AFAIK.

That would be OK as a temporary workaround while we figure out what to do with this.

> Opposed to re-adding it just for 1/5 users and 1/11 total keys.

I understand you see it as a burden, but right now it looks like the only remotely secure way to approach key management with sops.

adisbladis: I've added a software based key for now :/

zowoq:
I've reencrypted the secrets with the new key.

zimbatm (approved):
bors merge

.sops.yaml
zimbatm: My age key doesn't even have a passphrase because it's not supported very well by sops. https://github.com/getsops/sops/issues/933 . And GPG is quite archaic and has tons of UX problems.

It would be nice to find a solution that supports hardware dongles / 2FA

Mic92: @adisbladis if you use a commandline password manager you can also do:

```
export SOPS_AGE_KEY=$(pass some-key)
```

PR #650, opened 2023-07-01
modules: refactor common

zowoq:


zimbatm (approved):
I'm not opposed to it, but I don't understand the purpose of these small refactor PRs. Are they only stylistic, or are they done in order to unblock other changes?

zowoq:
Mostly stylistic but the intent is make modules sorted and consistent, bit easier to understand and work with. I should have done this originally rather than just moving our previous directory structure into `/modules`.

PR #652, opened 2023-07-02
refactor storage box

zowoq:
this is now under the nix-community hetzner account

### DIFF
--- a/hosts/build02/nixpkgs-update-backup.nix
+++ b/hosts/build02/nixpkgs-update-backup.nix
@@ -1,15 +1,8 @@
 { config, ... }:
 {
-  # 100GB storagebox is under the build03 hetzner account
+  # 100GB storagebox is under the nix-community hetzner account
 
-  sops.secrets.hetzner-build03-borgbackup-ssh = { };
-
-  programs.ssh.knownHosts = {
-    "hetzner-storage-box" = {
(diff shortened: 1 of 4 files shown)


hosts/build02/nixpkgs-update-backup.nix
```diff
@@ -23,7 +16,7 @@
     encryption.mode = "none";
     compression = "auto,zstd";
     startAt = "daily";
-    environment.BORG_RSH = "ssh -oPort=23 -i ${config.sops.secrets.hetzner-build03-borgbackup-ssh.path}";
+    environment.BORG_RSH = "ssh -oPort=23 -i ${config.sops.secrets.hetzner-borgbackup-ssh.path}";
     preHook = ''
       set -x
     '';
```
Mic92, line 22: Also made this part of the monitoring rules: https://github.com/Mic92/dotfiles/pull/714/files#diff-9275f84115cba354f44d705e21a10b1631cb34e64ff3377bbad9310af4a97acdR93

zowoq: What's the difference between the new rule and `daily_task_not_run` which has fired once for this?

Mic92: If the metric is completely absent, then Prometheus wouldn't knew about it anymore (i.e. there is no exporter providing old values)

PR #653, opened 2023-07-02
switch to mergify

zowoq:
this lets us consolidate automerges and remove GH_TOKEN_FOR_UPDATES

not using mergify merge queue function as we'd need to modify the effects.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,17 @@
+pull_request_rules:
+  - name: automatic merge on CI success
+    conditions:
+      - base=master
+      - check-success=ci/hercules/evaluation
+      - check-success=ci/hercules/onPush/default
+      - or:
+          - and:
+              - author=github-actions[bot]
+              - label=automerge
+          - and:
+              - author=hercules-ci[bot]
+    actions:
+      merge:
+        method: rebase
+        allow_merging_configuration_change: true
+      delete_head_branch: {}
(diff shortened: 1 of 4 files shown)


.github/workflows/flake-updates-nixpkgs-update.yml
```diff
@@ -13,12 +13,8 @@ jobs:
           extra_nix_config: |
             experimental-features = nix-command flakes
       - name: Update flake.lock
-        id: update
         uses: DeterminateSystems/update-flake-lock@v19
         with:
-          token: ${{ secrets.GH_TOKEN_FOR_UPDATES }}
```
Mic92, line 19: I think you still need the following permissions:

```
permissions:
  pull-requests: write
  contents: write
```

.github/mergify.yml
```diff
@@ -0,0 +1,17 @@
+pull_request_rules:
+  - name: automatic merge on CI success
+    conditions:
+      - base=master
+      - check-success=ci/hercules/evaluation
```
Mic92, line 5: If we ever have github actions than there are cases where github token is required to merge automatic pull requests. I think this is when an automatic PR is created and mergify does not need to merge any updates into it because it's up-to-date. In those cases github's CI is not triggered.

.github/workflows/flake-updates-nixpkgs-update.yml
zowoq: In this context the action has write so we don't need to set permissions unless we want to restrict write to specific steps.

Mic92: Ok. I only remember this being not the case in some orgs also I haven't check when this is the case.

zowoq:
Misunderstood how it works, doesn't currently work with github merge queue. https://github.com/nix-community/infra/pull/656

PR #654, opened 2023-07-02
secrets.yaml: refactor

zowoq:


secrets.yaml
```diff
@@ -1,11 +1,34 @@
...
-cloudflare.com: ENC[AES256_GCM,data:HtBSlzYtzeV4qrBlNdQXNQGRwEt/4dKt6iH4+ryOHib5HDJxJhG6zYneItFxQ10/fuJlXqLizfladsUzADvvEwCp4KJ06HaZn1do56J3lelDz+ROOuKzvTncGt/Fu4rjwT2SSHXmV+XTy8B3iBXDoX/d/u87J+j5y+BLY2XDLcq7YMDizyV/qg4kboKp9r5lravvNx2ZOj9T1TnKkJBJ32lH7Iz7SHo4qYOc/IBGXAS+e2qGVPCLWuygKg3mxTJ2Y/y6ba6JuC97fEo=,iv:dOACToYa8S44ANNIsHktD/lAG+VR/+mMqcdF8n77yaA=,tag:eqtUmKOGb/hFLIRl0FXNeA==,type:str]
-hetzner.com: ENC[AES256_GCM,data:rcYjYibR7lrdr8JTJ/ljS913kTjsC2J8HEcORA15x52l7+yYFS6FCsrWanFQaM8yc2w7NDp6+ND8PhQh5PLzuIJZyWHcosetJ9/Ow3SJjqNMnH2GogD/23iS3zwg2gKvnyMp/7gDIUkzsqHVysx3GIWjzq7Hdyq8Ls9/GIPqU8VYccR+nQ==,iv:kXygbJHHIJUkz/NAuAPBxqGAepLO/o3WVYz6YEwmrio=,tag:SKOCB1T8Cv5qkH856UAHvw==,type:str]
-darwin02-m1: ENC[AES256_GCM,data:0WKa4cWpEqWvZrEn,iv:ig+gD7ucGVh2WKVzmuxMiM9A2RJVM0E5avRzzc2Zqjk=,tag:6y0rEvjui0spqeE9uP+L7w==,type:str]
+build02:
+    root: ENC[AES256_GCM,data:XCyC99IsbFlMrL0iNazU,iv:CF+QthpBj4fJtUDLh4T+2ChDgfHKO1qZsO68usASIxA=,tag:83CGuJKxjsEu6DdkcN4QOQ==,type:str]
```
Mic92, line 2: Is this password still used? I don't see password hashes being set, but maybe we should.

zowoq: Don't think it's used, also current ssh settings don't allow root with a password.

Guess we can drop `root` for `build02` and `build04`?

> I don't see password hashes being set, but maybe we should.

Not sure why we'd want this?

Mic92: One can get a KVM console from hetzner for troubleshooting. There it might come in handy.

zowoq: I don't really see that we need it but I won't object if you want to set them.

Would you like to keep these current passwords or will you generate new passwords?

Mic92: You can drop this password for now. If I add new ones, I can do another pull request.

adisbladis (approved):

PR #661, opened 2023-07-07
terraform/cloudflare: fix build04-AAAA

zowoq:


zimbatm:
Is that normal to have an IPv6 with so many leading zeros on the WAN?

zowoq:
Yes? I don't see a problem with it but I'm not sure if I understand what you're asking?

Also noticed that the current `darwin02` is valid but doesn't seem to work with cloudflare.

terraform/cloudflare_nix-community_org.tf
```diff
@@ -59,7 +59,7 @@ resource "cloudflare_record" "nix-community-org-build04-A" {
 resource "cloudflare_record" "nix-community-org-build04-AAAA" {
   zone_id = local.nix_community_zone_id
   name    = "build04"
-  value   = "2603:c022:c001:b500:f1d4:5343:e8ce:d6ba"
+  value   = "0000:0000:0000:0000:0000:ffff:8d90:c91f"
```
Mic92, line 62: This is definitely not a valid public ipv6 address, whereas the old one is.

```
╭─[~]
╰─ % sipcalc 0000:0000:0000:0000:0000:ffff:8d90:c91f/64
-[ipv6 : 0000:0000:0000:0000:0000:ffff:8d90:c91f/64] - 0

[IPV6 INFO]
Expanded Address        - 0000:0000:0000:0000:0000:ffff:8d90:c91f
Compressed address      - ::ffff:8d90:c91f
Subnet prefix (masked)  - 0:0:0:0:0:0:0:0/64
Address ID (masked)     - 0:0:0:0:0:ffff:8d90:c91f/64
Prefix address          - ffff:ffff:ffff:ffff:0:0:0:0
Prefix length           - 64
Address type            - Reserved
Network range           - 0000:0000:0000:0000:0000:0000:0000:0000 -
                          0000:0000:0000:0000:ffff:ffff:ffff:ffff

-

╭─[~]
╰─ % sipcalc 2603:c022:c001:b500:f1d4:5343:e8ce:d6ba
-[ipv6 : 2603:c022:c001:b500:f1d4:5343:e8ce:d6ba] - 0

[IPV6 INFO]
Expanded Address        - 2603:c022:c001:b500:f1d4:5343:e8ce:d6ba
Compressed address      - 2603:c022:c001:b500:f1d4:5343:e8ce:d6ba
Subnet prefix (masked)  - 2603:c022:c001:b500:f1d4:5343:e8ce:d6ba/128
Address ID (masked)     - 0:0:0:0:0:0:0:0/128
Prefix address          - ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff
Prefix length           - 128
Address type            - Aggregatable Global Unicast Addresses
Network range           - 2603:c022:c001:b500:f1d4:5343:e8ce:d6ba -
                          2603:c022:c001:b500:f1d4:5343:e8ce:d6ba
```

Mic92: It looks like the ipv6 address on build04 just disappeared. Maybe an issue with dhcpv6?

Mic92: According to oracle we should get 2603:c022:c001:b500:0000:0000:0000:0000/64 as the subnet.
Will check why we don't get that.

https://dl.thalheim.io/vLV19A1wYwLQzoUdDAZkzw/20230708-075855.png

Mic92: I can see ipv6 router advertisements being done with the managed flag (this is what I would expect for dhcpv6. However networkd does not issue dhcpv6 as a follow up and only adds the router / subnet:

```
2603:c022:c001:b500::/64 dev enp0s3 proto ra metric 1024 pref medium
fe80::/64 dev enp0s3 proto kernel metric 256 pref medium
default via fe80::200:17ff:fe7c:ce5b dev enp0s3 proto ra metric 1024 expires 8993sec pref medium
```

Mic92: Ok. Also dhcpcd fails to get an ipv6 address:

```
enp0s3: Router Advertisement from fe80::200:17ff:fe7c:ce5b
enp0s3: no global addresses for default route
enp0s3: adding route to 2603:c022:c001:b500::/64
Dropped protocol specifier '.ra' from 'enp0s3.ra'. Using 'enp0s3' (ifindex=2).
enp0s3: soliciting a DHCPv6 lease
enp0s3: no global addresses for default route
```

Mic92:
Here is the actual fix for build04: https://github.com/nix-community/infra/pull/663
ipv6 works again.

PR #662, opened 2023-07-08
terraform/cloudflare: fix darwin02-AAAA

zowoq:


Mic92 (approved):
Looks like a valid ipv6 address, however I seem to be not able to login also ssh connection succeeds.

zowoq:
We don't have separate accounts on this host, just added our keys under the default admin account, does `m1` user work?

Mic92:
Yes. That works.

PR #669, opened 2023-07-15
flake: set pkgs instead of system

zowoq:


flake.nix
```diff
@@ -118,23 +116,18 @@
...
             };
             build03 = nixosSystem {
-              system = "x86_64-linux";
               modules = [ ./hosts/build03/configuration.nix ];
             };
             build04 = nixosSystem {
-              system = "aarch64-linux";
               modules = [ ./hosts/build04/configuration.nix ];
             };
             web01 = nixosSystem {
```
Mic92, line 130: I believe we save evaluation time if we set pkgs directly because then nixpkgs is not re-imported for every system: https://github.com/Mic92/dotfiles/blob/05c0dbc5ba882fbe4d51cffe6e124cdfcd924983/nixos/flake-module.nix#L118

Mic92: Would have to benchmark that but if I had to guess a few 100MB per machine.

Mic92: Not so sure how much we safe for the infra repository:

https://github.com/TUM-DSE/doctor-cluster-config/pull/322

This repo has significantly more hosts.

zowoq: Everything is buildable but hercules doesn't seem to like `pkgs`, is there something else we need to do for this?

zowoq: @Mic92 Any suggestions for fixing this?

Mic92: Looks like it's fixed? Maybe unrelated?

Mic92:
ok. no, it came back.

flake.nix
Mic92: ok. no, it came back.

zowoq:
I think this needs to be fixed in hercules, I'll take a look once we have darwin CI working again.

Mic92:
Might be not worth the hassle

zowoq:
We should be able to merge this when this repo is using buildbot for CI.

zowoq:
Not sure why but it works now?

zowoq:
Seem to have changed to pre-git on darwin but doesn't really matter.

```
darwin-system  23.11.20231019.7c9cc5a+darwin4.8b6ea26 -> 23.11pre-git+darwin4.8b6ea26
```

PR #672, opened 2023-07-16
Revert "flake: pin nixpkgs"

zowoq:


zowoq:
formatter.x86_64-darwin is broken via prettier/nodejs but the other four darwin attributes are built and cached?

zowoq:
Deployed to darwin hosts.

PR #683, opened 2023-07-19
modules/darwin/telegraf: add swap

zowoq:


modules/darwin/common/telegraf.nix
```diff
@@ -7,6 +7,7 @@
         prometheus.metric_version = 2;
         system = { };
         mem = { };
+        swap = { };
```
Mic92, line 10: Do you want me to add checks for swap?

Mic92: Ah. I got one.

modules/darwin/common/telegraf.nix
```diff
@@ -7,6 +7,7 @@
...
         disk.tagdrop = {
           fstype = [ "tmpfs" "ramfs" "devtmpfs" "devfs" "iso9660" "overlay" "aufs" "squashfs" ];
```
Mic92, line 12: Pretty sure most of them don't exist on macos i.e. overlay aufs squashfs

modules/darwin/common/telegraf.nix
```diff
@@ -7,6 +7,7 @@
...
           device = [ "rpc_pipefs" "lxcfs" "nsfs" "borgfs" ];
```
Mic92, line 13: Maybe borgfs exists? The rest probably doesn't.

modules/darwin/common/telegraf.nix
zowoq: Yes, I just copied `disk.tagdrop` from the srvos module so it's easier to keep them in sync.

zowoq:
Deployed to darwin hosts.

PR #684, opened 2023-07-20
.sops.yaml: update path_regex for modules

zowoq:


.sops.yaml
```diff
@@ -74,7 +74,7 @@ creation_rules:
           - *zimbatm
           - *zowoq
           - *adisbladis
-  - path_regex: modules/darwin/hercules-ci/.+\.yaml$
+  - path_regex: modules/darwin/.+\.yaml$
```
Mic92, line 77: Time to add sops-nix support for darwin :) ? So far we only support home-manager on this target.

zowoq: > sops-nix support for darwin

Are there any technical issues that need to overcome for that or would it just be a case of adapting the darwin home-manager support for nix-darwin?

Mic92: I would expect it's more a case of adapting it based on what we do for home-manager.

PR #690, opened 2023-07-21
modules/darwin: add reboot

zowoq:


zowoq:
Forgot about this.

modules/darwin/common/reboot.nix
```diff
@@ -0,0 +1,13 @@
+{
+  # reboot every sunday between 00:00 and 06:00
+  launchd.daemons.reboot = {
+    command = ''/sbin/shutdown -r "+$(( $RANDOM % ( 6 * 60 ) ))"'';
```
Mic92, line 4: System upgrades have to performed manually, right?
I wonder how much value there is in rebooting except for resetting some counters.

zowoq: > System upgrades have to performed manually, right?

Yeah, I've been doing the system and security upgrades manually. Wouldn't want to just automatically upgrade everything unconditionally and at this scale probably isn't worth the hassle of trying to manage it programmatically. Might look at getting the updates into the community monitoring room for visibility, either from the machines themselves or a feed from apple.

> I wonder how much value there is in rebooting except for resetting some counters.

I haven't found macos to be very good for extended uptime, after a couple of weeks something starts causing a problem, easier to just reboot weekly and not worry about it.

Mic92: Telegraf allows for integration with custom scripts fairly easily.

PR #708, opened 2023-07-28
docs: fix OpenCollective banner height

zimbatm:
I couldn't find a way to set the height in their docs, so hack it.

### DIFF
--- a/docs/open-collective.md
+++ b/docs/open-collective.md
@@ -1,1 +1,6 @@
+<style>
+div.opencollective-banner > iframe {
+    height: 700px;
+}
+</style>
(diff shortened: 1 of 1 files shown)


dev/shell.nix
```diff
@@ -6,6 +6,7 @@
         jq
         python3.pkgs.deploykit
         python3.pkgs.invoke
+        python3.pkgs.mkdocs-material
```
zowoq, line 9: Removed this in https://github.com/nix-community/infra/commit/486311ad294ec325796f99e78bccbb7b991e261b to reduce bloat, can build docs locally with `inv mkdocs`.

zimbatm: Removed

PR #711, opened 2023-07-29
modules/nixos: monitoring, web02: monitoring

zowoq:


modules/nixos/monitoring/telegraf.nix
```diff
@@ -0,0 +1,43 @@
+{
+  services.telegraf.extraConfig.inputs = {
```
zowoq, line 2: @Mic92 I was going to move this stuff first so the following PR can focus on prometheus, etc.

PR #722, opened 2023-07-31
modules/darwin/common: disable alf

zowoq:


modules/darwin/common/default.nix
```diff
@@ -36,6 +36,9 @@
     htop
   ];
 
+  # disable application layer firewall, telegraf needs an incoming connection
+  system.defaults.alf.globalstate = 0;
```
zowoq, line 40: This was already disabled on the scaleway machine when I was setting it up but it was enabled on the hetzner machines.

PR #723, opened 2023-08-01
build05: init

zowoq:


devdoc/hosts.md
```diff
@@ -37,6 +37,16 @@ This machine is meant as an aarch64 builder for our hydra instance running on bu
 - RAM: 24GB
 - Drives: 200 GB Block
 
+### `build05`
+
+This machine is meant as an aarch64 builder for our CI.
+
+- Provider: Hetzner
+- Instance type: [CAX41](https://www.hetzner.com/cloud)
```
zowoq, line 45: @Mic92 Do you already have any config for this type? What filesystem should we use?

Mic92: https://github.com/numtide/srvos/pull/214

Mic92: Added disko config.

Mic92: Question: Why do we need another ci builder for aarch64 that is also a VM? Originally we wanted to buy a bare-metal builder so that we can run VM tests. But it seems like another aarch64 vm is not needed unless we run into capacity issues or oracle decides kick us out.

zowoq: Replied on matrix.

PR #727, opened 2023-08-04
add Namespace to the project

zimbatm:
Thanks for the free runners!


zowoq:
Can we link the image like we do for the other sponsors?

Mic92:
The only logo I found is this one: https://github.com/namespacelabs/foundation/blob/c13724b4aabdbf3f782569107ff1317cd06acb58/internal/webui/devui/public/icons/logo_filled.svg?plain=1 But it does not have the name in it. Linking to some website seems a bit risky since the path might change.

PR #735, opened 2023-08-07
modules/nixos/community-builder: remove trusted from users

zowoq:


zowoq:
I'll leave this until the default `nix` version in nixpkgs is > 2.15.

Will ping everyone who has trusted when this is merged.

zowoq:
@0x4A6F
@a-kenji
@fgaz
@flokli
@Janik-Haag
@jtojnar
@nlewo
@lilyinstarlight
@raitobezarius
@schmittlauch

We'd like to test removing `trusted` access for all users on this machine, nix 2.16 supports "Trustless remote building for input-addressed drvs". https://github.com/NixOS/nix/issues/2789, https://github.com/NixOS/nix/pull/3921

Is not having `trusted` access or using a nix version >= 2.16 a problem for anyone?

fgaz:
I often use `trusted` access to run nixos tests

zowoq:
> I often use `trusted` access to run nixos tests

I think that with nix 2.16 you should be able to do that without trusted?

RaitoBezarius:
As Nix stable on unstable is 2.17, thanks to you, I guess, this is not a
big deal IMHO.

Though if we discover bugs and we had to revert temporarily to a working
version of Nix, I think it can be a bit annoying, generally, it happens to
me some time, so if we turn it on, let's keep the option to disable it
again if there's bugs in Nix ≥ 2.17 that cannot be worked around in a
reasonable time.

Thanks!
-- 

Ryan Lahfa

fgaz:
> I think that with nix 2.16 you should be able to do that without trusted?

then it's not a problem for me

schmittlauch:
While I am continuing to use NixOS stable and thus am on nix-2.13, I do not regularly use the community builder right now. So this doesn't immediately break things, just thanks for the heads up.

zowoq:
If this does cause any problems you can let us know here or on #nix-community matrix.

zowoq:
We've reverted this so now everyone is trusted: https://github.com/nix-community/infra/pull/962.

PR #736, opened 2023-08-07
build01: add hercules and remote builder

zowoq:
depends on https://github.com/nix-community/infra/pull/735

If we merge this we could remove hercules from build02 and try increasing the number of nixpkgs-update workers again.


Mic92:
If we have this we shouldn't enable trusted builder again because it could be used to push to the nix binary cache.

zowoq:
Closing for now, might revisit if we can figure out a way of prioritizing community builds over CI jobs.

PR #753, opened 2023-08-12
modules/nixos/monitoring: add grafana

zowoq:


modules/nixos/monitoring/grafana.nix
```diff
@@ -0,0 +1,38 @@
+{ config, ... }:
+{
+  services.grafana = {
+    enable = true;
+    settings = {
+      analytics.reporting_enabled = false;
+
+      "auth.anonymous".enabled = true;
```
Mic92, line 8: We could use the github org for authentication here. Than we don't need to create accounts manually.

modules/nixos/monitoring/grafana.nix
```diff
@@ -0,0 +1,103 @@
...
+    ensureDatabases = [ "grafana" ];
+    ensureUsers = [
+      {
+        name = "grafana";
+        ensurePermissions = { "DATABASE grafana" = "ALL PRIVILEGES"; };
+      }
+    ];
+  };
+
+  services.postgresqlBackup = {
```
zowoq, line 58: I'll open another PR for backups after this is merged.

modules/nixos/monitoring/grafana.nix
```diff
@@ -0,0 +1,69 @@
...
+{
+  systemd.services.grafana.after = [ "prometheus.service" ];
+
+  services.grafana = {
+    enable = true;
+    settings = {
+      analytics.reporting_enabled = false;
+      analytics.feedback_links_enabled = false;
+
+      "auth.anonymous".enabled = false;
```
zowoq, line 11: Don't see much point in making this public, only really of interest to nix-community org members.

zowoq: We will actually want this public for logs from nixpkgs-updates with loki/promtail so I'll just leave it enabled.

PR #758, opened 2023-08-14
modules/nixos/monitoring/alert-rules: add matrix-alertmanager-receiver

zowoq:
Separate rule for the matrix service to improve visibility.

### DIFF
--- a/modules/nixos/monitoring/alert-rules.nix
+++ b/modules/nixos/monitoring/alert-rules.nix
@@ -89,6 +89,11 @@ lib.mapAttrsToList
       description = "{{$labels.host}} failed to (re)start service {{$labels.name}}";
     };
 
+    matrix_alertmanager_receiver_not_running = {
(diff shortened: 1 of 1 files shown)


modules/nixos/monitoring/alert-rules.nix
```diff
@@ -89,6 +89,11 @@ lib.mapAttrsToList
       description = "{{$labels.host}} failed to (re)start service {{$labels.name}}";
     };
 
+    matrix_alertmanager_receiver_not_running = {
+      condition = ''systemd_units_active_code{name="matrix-alertmanager-receiver.service", instance="localhost:9273", sub!="running"}'';
```
Mic92, line 93: I think you can safely assume that a matrix-alertmanager-receiver should run always no matter on which host:

```suggestion
      condition = ''systemd_units_active_code{name="matrix-alertmanager-receiver.service", sub!="running"}'';
```

PR #760, opened 2023-08-14
modules/nixos/monitoring/prometheus: drop `match_re`

zowoq:


modules/nixos/monitoring/prometheus.nix
```diff
@@ -78,7 +77,6 @@
           webhook_configs = [
             {
               url = "http://localhost:9088/alert";
-              max_alerts = 5;
```
zowoq, line 81: @Mic92 I think this is okay to drop?

Mic92: I mainly had it in there to not disturb conversations if we get a lot of alerts at once.

PR #761, opened 2023-08-14
modules/nixos/monitoring: various

zowoq:


modules/nixos/monitoring/alert-rules.nix
```diff
@@ -101,8 +101,13 @@ lib.mapAttrsToList
       description = "{{$labels.host}} is running with load15 > 1 for at least 5 minutes: {{$value}}";
     };
 
+    localhost_reboot = {
```
zowoq, line 104: This is to ensure that we notice if the matrix alerts stop working.

This machine will reboot once or twice a week after a flake update and send an alert.

modules/nixos/monitoring/prometheus.nix
```diff
@@ -48,6 +41,11 @@
     ];
   };
 
+  services.telegraf.extraConfig.inputs.prometheus.urls = [
+    "http://localhost:9090/metrics" # prometheus
+    "http://localhost:9093/metrics" # alertmanager
```
zowoq, line 46: @Mic92 Do you have something like this in your config and I've overlooked it?

I don't think rules like `alerts_silences_changed`, `prometheus_notifications_alertmanagers_discovered` will work without it?

zowoq: Merging this PR but if you think this should be handled differently let me know and I'll change it.

PR #780, opened 2023-08-25
community-builder packages: Add mosh

matthiasbeyer:
I would like to propose to add `mosh` ([mosh.org](https://mosh.org/)) to the community builder.

I like to keep my ssh sessions via suspends, and this helps.

### DIFF
--- a/modules/nixos/community-builder/packages.nix
+++ b/modules/nixos/community-builder/packages.nix
@@ -18,4 +18,12 @@
     pkgs.termite.terminfo
     pkgs.wezterm.terminfo
   ];
+
+  networking.firewall.allowedUDPPortRanges = [
+    # Reserved for mosh
+    {
+      from = 60000;
+      to = 61000;
+    }
+  ];
 }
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
Ah, well... this would also require some ports to be opened... and you might not want that?

modules/nixos/community-builder/packages.nix
```diff
@@ -4,6 +4,7 @@
   environment.systemPackages = [
     pkgs.fd
     pkgs.git
+    pkgs.mosh
```
Mic92, line 7: Should be fine:

```
  networking.firewall.allowedUDPPortRanges = [
    # Mosh
    {
      from = 60000;
      to = 60010;
    }
  ];
```

matthiasbeyer:
Added a patch for the UDP port range required for mosh.

1. This should probably go in another file, please point me to it
2. I added the range 60000-61000, not as @Mic92 requested 60000-60010 - I added the range that mosh might use. Is that okay or shall I restrict the range to 10/11 ports?

zowoq:
> This should probably go in another file, please point me to it

This file is fine.

PR #785, opened 2023-08-28
refactor mkdocs

zowoq:
- use devshell instead of package

- build pages with hercules-ci.github-pages

- drop unnecessary checks restriction


zimbatm:
What problem does this change resolve? The package was helping verify that the docs will keep working before the deployment

zowoq:
> The package was helping verify that the docs will keep working before the deployment

`checks.github-pages-effect-is-buildable` does this.

zimbatm (approved):

PR #786, opened 2023-08-28
modules/nixos/monitoring: set after for services

zowoq:


zimbatm:
Is that change related to a problem that you observed?

zowoq:
Yes, alertmanager failed after the last reboot.

PR #793, opened 2023-08-31
.github/workflows/fork.yml: fake hercules status for merge queue

zowoq:
Hopefully this will allow us to merge PRs from forks.

Tested in https://github.com/nix-community/infra/pull/794.

### DIFF
--- a/.github/workflows/fork.yml
+++ b/.github/workflows/fork.yml
@@ -1,0 +1,18 @@
+name: "hercules"
+
+on:
+  pull_request_target:
+
+jobs:
+  eval:
+    name: ci/hercules/evaluation
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/fork.yml
```diff
@@ -0,0 +1,18 @@
+name: "hercules"
+
+on:
+  pull_request:
```
zowoq, line 4: This actions isn't sensitive so I think `pull_request` should be fine.

Mic92: It's possible to make the approval optional afaik via project settings.

zowoq: Probably easier to keep this as pull_request_target like the label action.

zowoq:
It does go through hercules via the merge queue but it isn't shown in the PR. 

ci

PR #797, opened 2023-08-31
build02/nixpkgs-update: tee worker stdout and stderr to log files

Mic92:


Mic92:
Why was automerge disabled?

rhendric (approved):
I don't know if you added the mkdir because you tested this and it was necessary, or if it was something you thought would be necessary; if the latter, as mentioned in the other PR I think the `-D` option to rotatelogs creates parent directories as needed. But the mkdir doesn't hurt in any case.

Mic92:
@zowoq ok. I leave this to pull request to you than. Knowing @ryantm, I was expecting him to be fine with this, but he sometimes doesn't have time to look into large amounts of pull request done in this repository.

hosts/build02/nixpkgs-update.nix
```diff
@@ -47,6 +48,10 @@ let
     };
 
     script = ''
+      mkdir -p "$LOGS_DIRECTORY/~workers/"
```
Mic92, line 51:
```suggestion
      mkdir -p "$LOGS_DIRECTORY/~workers/"
      # This is for public logs at https://r.ryantm.com/log/~workers
```

ryantm (approved):

PR #812, opened 2023-09-05
modules/shared/nix-daemon: fix `asGB`

zowoq:
min-free, max-free and max-freed numbers are from https://github.com/NixOS/nixos-org-configurations/blob/master/macs/nix-darwin.nix, I think they're fine for all the hosts.

### DIFF
--- a/modules/shared/nix-daemon.nix
+++ b/modules/shared/nix-daemon.nix
@@ -1,7 +1,7 @@
 { pkgs, ... }:
 
 let
-  asGB = size: toString (size * 1024 * 1024);
+  asGB = size: toString (size * 1024 * 1024 * 1024);
 in
 {
   nix = {


Mic92:
Do we have a raid0 already on those builders? That might also increase cache size.

zowoq:
No, the hetzner machines are still zfs mirrors. 

I'll probably have another go at converting them at some point, the disko/nixos-anywhere vm tests should make it a bit easier to test.

Mic92 (approved):

zowoq:
Changed this to keep the original gc settings for the `web` hosts.

zowoq:
This gc settings are still a bit awkward, I'll merge min/max free here and continue with the gc settings in https://github.com/nix-community/infra/pull/814.

PR #833, opened 2023-09-14
build02/nixpkgs-update: replace fifo with supervisor service

### LINKED ISSUE #799: nixpkgs-update workers are occasionally eating garbage
The FIFO used to communicate between fetchers and workers is not a safe replacement for a proper task queue.

Excerpts from https://r.ryantm.com/log/~workers/2023-08-31-worker1.stderr.log:
```
+ /var/lib/nixpkgs-update/bin/nixpkgs-update update-batch --pr --outpaths --nixpkgs-review 'rg.r/rjc/tlasle/esoshaskellPackages.llvm-dsl 0.0 0.1 https://repology.org/project/haskell:llvm-dsl/versions'
[...]
+ /var/lib/nixpkgs-update/bin/nixpkgs-update update-batch --pr --outpaths --nixpkgs-review 'ython31akgsti-esce .0201. tp:/eooyogpoetpto:rowboktvrin'
```
And from https://r.ryantm.com/log/~workers/2023-08-31-worker2.stderr.log:
```
+ /var/lib/nixpkgs-update/bin/nixpkgs-update update-batch --pr --outpaths --nixpkgs-review 'p1Pcae.rowbokt01. .03hts/rplg.r/rjc/yhnti-esce/esoskabeljau 1.2.0 2.0.0 https://repology.org/project/kabeljau/versions'
```

I want to replace the FIFO with a queue service using sockets, and make the queue a least-recently-updated priority queue while I'm at it; this issue is just advance warning of my intentions, in case they interfere with someone else's.

rhendric:
Fixes #799.

I went to some lengths to build the entire `build02` machine as a container and run it to make sure that this works, but between not knowing how to stub in my own values for secrets and my suspicion that production `build02` has retained some state that the configuration doesn't create (notably, `/var/cache/nixpkgs-update/fetcher/nixpkgs`), the best I was able to do is verify that this all works once I kick the fetchers manually.

High-level description of changes:
 * Created a new service called `nixpkgs-update-supervisor` to replace the named pipe previously used for communication between fetchers and workers. The service implements a priority queue so that no package goes unserviced for too long. The queue is persisted across runs in a SQLite database accessible from https://r.ryantm.com/log/~supervisor, so it can be inspected to get statistics or help diagnose systemic problems.
   * Added a unit test suite for the supervisor script. This isn't automated or hooked up to anything. I wrote it for myself and thought the next person to work on this service might benefit from it.
 * Fetchers are rewritten to write their output to normal files, so that their output is also persisted between runs. These files will also be accessible from https://r.ryantm.com/log/~fetchers.
 * Fetcher services restart themselves after 24 hours. I have no idea how fetchers were previously being re-run—did someone trigger them manually? Did they just get restarted every time the infrastructure was deployed and that was frequent enough?
 * Fetchers and workers also restart `on-failure` instead of just `on-abort`, which covers things like connection problems between fetcher and GitHub or Repology.
 * Workers do a check when they receive a job to see if their copy of `nixpkgs` exists yet. If it doesn't, one of the workers takes a lock while the others block. This is because `nixpkgs-update` isn't that smart about cloning Nixpkgs; if one worker is cloning and a second worker tries to run, it will fail almost immediately. That would result in a lot of jobs being pulled from the queue and then thrown away. This is unlikely to matter in production since the workers' copy of `nixpkgs` seems to be another piece of persistent state that make production different from a test environment, but again I wrote it to make my life easier and I thought it might be helpful to someone else too.

### DIFF
--- a/hosts/build02/nixpkgs-update.nix
+++ b/hosts/build02/nixpkgs-update.nix
@@ -14,12 +14,13 @@ let
     getent # used by hub
     cachix
     apacheHttpd # for rotatelogs, used by worker script
+    socat # used by worker script
   ];
 
   nixpkgs-update-github-releases' = "${inputs.nixpkgs-update-github-releases}/main.py";
 
   mkWorker = name: {
-    after = [ "network-online.target" ];
+    after = [ "network-online.target" "nixpkgs-update-supervisor.service" ];
     wantedBy = [ "multi-user.target" ];
     description = "nixpkgs-update ${name} service";
     enable = true;
@@ -53,25 +54,53 @@ let
       exec  > >(rotatelogs -eD "$LOGS_DIRECTORY"'/~workers/%Y-%m-%d-${name}.stdout.log' 86400)
       exec 2> >(rotatelogs -eD "$LOGS_DIRECTORY"'/~workers/%Y-%m-%d-${name}.stderr.log' 86400 >&2)
 
-      pipe=/var/lib/nixpkgs-update/fifo
+      socket=/run/nixpkgs-update-supervisor/work.sock
 
-      if [[ ! -p $pipe ]]; then
-        mkfifo $pipe || true
-      fi
+      function run-nixpkgs-update {
+        exit_code=0
+        set -x
+        ${nixpkgs-update-bin} update-batch --pr --outpaths --nixpkgs-review "$attr_path $payload" || exit_code=$?
+        set +x
+        msg="DONE $attr_path $exit_code"
+      }
 
-      exec 8<$pipe
-      while true
-      do
-        if read -u 8 line; then
-          set -x
-          ${nixpkgs-update-bin} update-batch --pr --outpaths --nixpkgs-review "$line" || true
-          set +x
-        fi
+      msg=READY
+      while true; do
+        response=$(echo "$msg" | socat UNIX-CONNECT:"$socket" - || true)
+        case "$response" in
+          "") # connection error; retry
+            sleep 5
+            ;;
+          NOJOBS)
+            msg=READY
+            sleep 60
+            ;;
+          JOB\ *)
+            read -r attr_path payload <<< "''${response#JOB }"
+            # If one worker is initializing the nixpkgs clone, the other will
+            # try to use the incomplete clone, consuming a bunch of jobs and
+            # throwing them away. So we use a crude locking mechanism to
+            # run only one worker when there isn't a nixpkgs directory yet.
+            # Once the directory exists and this initial lock is released,
+            # multiple workers can run concurrently.
+            lockdir="$XDG_CACHE_HOME/.nixpkgs.lock"
+            if [ ! -e "$XDG_CACHE_HOME/nixpkgs" ] && mkdir "$lockdir"; then
+              trap 'rmdir "$lockdir"' EXIT
+              run-nixpkgs-update
+              rmdir "$lockdir"
+              trap - EXIT
+              continue
+            fi
+            while [ -e "$lockdir" ]; do
+              sleep 10
+            done
+            run-nixpkgs-update
+        esac
       done
     '';
   };
 
-  mkFetcher = cmd: {
+  mkFetcher = name: cmd: {
     after = [ "network-online.target" ];
     wantedBy = [ "multi-user.target" ];
     path = nixpkgsUpdateSystemDependencies;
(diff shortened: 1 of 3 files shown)


Mic92:
Wow. You seem to have put a lot of work into this change. If you haven't done so, I would try to get in touch with @ryantm. Usually it's a good idea to get feedback on things like that before doing these big refactorings so that in the end both sides are happy with the change. I remember that ryan was unhappy with the way builds are retried so far so that might be a welcome change?

Mic92:
> Need to fix python formatting and linting before merging otherwise the treefmt check will block this.

You can do this by running `nix fmt` locally.

rhendric:
> You can do this by running `nix fmt` locally.

Thanks, I wasn't familiar with that command.

ryantm (approved):
I looked it over. Let's try it out! Maybe we can get @rhendric ssh/sudo access so they can debug as needed? Maybe at least temporarily, if permanently is a problem.

zowoq:
>  Maybe we can get rhendric ssh/sudo access so they can debug as needed? Maybe at least temporarily, if permanently is a problem.

Permanent sudo would mean that we're basically adding another infra admin. Even temporarily isn't easy as sudo gives access to secrets on the box, some of which are shared.

I could remove all of the non-`nixpkgs-update` secrets/services from this box for now and re-add them later when this is finished?

ryantm:
Okay, I'll try to deploy it myself soon!

zowoq:
I've made a separate PR for the `on-failure` changes, should reduce the amount of failed alerts that we're getting from these services. https://github.com/nix-community/infra/pull/873

ryantm:
Finally got some time. Deployed this now and am monitoring.

rhendric:
Well, this is useful!

Look at the updatescript.*.txt file in https://r.ryantm.com/log/~fetchers/.

Notice how it still has `python39Packages` entries and no `python311Packages` entries. This is clearly generated from a rather old copy of Nixpkgs, which goes a long way toward explaining why some packages with update scripts haven't been getting attention from r-ryantm.

The path being used by `packages-with-update-script.nix` is `/var/cache/nixpkgs-update/updatescript/nixpkgs`; I didn't see anything else in the build02 configuration mentioning this path. Should `packages-with-update-script.nix` point somewhere else, or is it intended that this fetcher has its own copy of Nixpkgs and we need some machinery to periodically pull it?

ryantm:
This PR seems to be working great. Thank you @rhendric!

I've manually updated /var/cache/nixpkgs-update/updatescript/nixpkgs for now. I'll try to look into it more.

ryantm:
https://github.com/nix-community/infra/pull/878

hosts/build02/nixpkgs-update.nix
```diff
@@ -86,21 +115,26 @@ let
...
       CacheDirectoryMode = "700";
     };
-    script = ''
-      pipe=/var/lib/nixpkgs-update/fifo
-
-      if [[ ! -p $pipe ]]; then
-        mkfifo $pipe || true
-      fi
 
-      ${cmd} | sort -R > $pipe
```
zowoq, line 103: Could we restore randomisation? Two scenarios I think it would help with:

Duplicate PRs (same version bump, different attribute name) opened by separate workers at the same time.

Sometimes going multiple hours without opening a PR when it working through package sets. I'm reluctant to expand the bot skiplist or start putting `DO NOT UPDATE` on stuff in nixpkgs as sometimes it just seems to be transient breakage and also having recent failure logs could be helpful in fixing actual problems.

rhendric: I think we could do something that fuzzes the initial order in which newly ingested packages are queued up. After the first full go-around, existing packages will be dispatched in roughly the same order every time, but for the things you're pointing out I don't think that should matter, yeah?

zowoq: Yes, that should be fine. Thanks for implementing it.

zowoq: Not sure what happened here, submitted four PRs within a few minutes for the same vaultwarden update.

vaultwarden

rhendric: It's possible there's a bug, but it's also possible for this to happen organically as designed. If a fetcher emits any never-before-seen attribute paths, the supervisor still prioritizes those over attribute paths it has attempted to update before. If there are only a small number of such paths for a given run of a fetcher, they're going to get bunched together.

PR #868, opened 2023-09-26
modules/nixos/monitoring: add loki, modules/nixos/common: add promtail

zowoq:


modules/nixos/monitoring/loki.nix
```diff
@@ -0,0 +1,15 @@
+{
+  services.telegraf.extraConfig.inputs.prometheus.urls = [
+    "http://localhost:3100/metrics"
+  ];
+
+  services.loki = {
```
Mic92, line 6: It took me a while to find sane loki configuration:

https://github.com/Mic92/dotfiles/blob/c912c73266f10973106a8522cbec827e2802918d/nixos/eva/modules/loki.nix#L39

zowoq:
We've been doing okay without this so far and at the moment it seems like it may not be worth the effort, may need to reconsider if we end up adding more hosts or we start buying our own hardware.

zowoq:
> We've been doing okay without this so far and at the moment it seems like it may not be worth the effort, may need to reconsider if we end up adding more hosts or we start buying our own hardware.

Still seems to be true.

PR #876, opened 2023-10-01
use `sotp` for TOTP secrets

zowoq:
I'd imagine that eventually we'd want to move to a password manager but this is enough of an improvement over the current workflow that we can put off making that decision.

https://github.com/getsops/sotp

Unlicensed at the moment, I'll add it to nixpkgs once upstream addresses that.

### DIFF
--- a/dev/shell.nix
+++ b/dev/shell.nix
@@ -15,5 +15,22 @@
         python3.pkgs.mkdocs-material
       ];
     };
+    sotp = with pkgs; mkShellNoCC {
+      packages = [
+        (buildGoModule rec {
+          pname = "sotp";
+          version = "e7f7c804b1641169ce850d8352fb07294881609e";
+          src = pkgs.fetchFromGitHub {
+            owner = "getsops";
+            repo = "sotp";
+            rev = version;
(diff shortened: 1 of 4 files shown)


Mic92:
What shared totp secrets do we currently have?

zimbatm:
the github bot accounts for example, since GitHub requires MFA on all the accounts now

Mic92 (approved):

dev/shell.nix
```diff
@@ -15,5 +15,22 @@
         python3.pkgs.mkdocs-material
       ];
     };
+    sotp = with pkgs; mkShellNoCC {
```
zowoq, line 18: Don't need this in the default devshell.

tasks.py
```diff
@@ -41,6 +41,14 @@ def deploy(h: DeployHost) -> None:
     g.run_function(deploy)
 
 
+@task
+def sotp(c: Any, acct: str) -> None:
+    """
+    Get TOTP token from sops
+    """
+    c.run(f"nix develop .#sotp -c sotp {acct}")
```
zimbatm, line 49: I think you could get away with making this nix runnable no? `f"nix run .#sotp {acct}"` (and change the above recipe)

zowoq: Once this package is added to nixpkgs we'd be changing it back to a devshell anyway?

PR #878, opened 2023-10-08
build02: nixpkgs-update: before running the updatescript fetcher, pull latest nixpkgs

ryantm:
Why
===
* The updatescript was not getting the latest information

What changed
===
* Before finding updatescripts to run, update nixpkgs


rhendric:
Need to silence the git command; see updatescript output in https://r.ryantm.com/log/~fetchers/

PR #885, opened 2023-10-14
terraform/shell: remove meta from terraform

zowoq:


flake.nix
```diff
@@ -52,11 +52,16 @@
           inputs.treefmt-nix.flakeModule
         ];
 
-        perSystem = { config, pkgs, ... }:
+        perSystem = { config, pkgs, system, ... }:
           let
             defaultPlatform = pkgs.stdenv.hostPlatform.system == "x86_64-linux";
           in
           {
+            _module.args.tf-pkgs = import inputs.tf-pkgs {
```
Mic92, line 60: No strong opinion but this variant is a bit less code and doesn't require a terraform nixpkgs instantiation: https://github.com/numtide/treefmt/commit/de751ab74117c7686493ee696c34d9e5d4e0ba11

PR #894, opened 2023-10-25
add buildbot

zowoq:


Mic92:
Note: For buildbot-nix, we also currently need https://github.com/Mic92/nixpkgs/commit/3bb2151753ed7f1c4a29e2d5983f626b3c4a6107 because of a recent change in systemd's LoadCredential.

Mic92:
Pull request is here: https://github.com/buildbot/buildbot/pull/7159

zowoq:
> Note: For buildbot-nix, we also currently need [Mic92/nixpkgs@3bb2151](https://github.com/Mic92/nixpkgs/commit/3bb2151753ed7f1c4a29e2d5983f626b3c4a6107) because of a recent change in systemd's LoadCredential.

We discussed this on matrix, you were going to move the patch to https://github.com/Mic92/buildbot-nix?

Mic92:
We potentially have to rework the permission system.

THis is what buildbot-nix does at the moment:
```
        config["www"]["authz"] = util.Authz(
            roleMatchers=[
                util.RolesFromUsername(roles=["admin"], usernames=self.github.admins)
            ],
            allowRules=[
                util.AnyEndpointMatcher(role="admin", defaultDeny=False),
                util.AnyControlEndpointMatcher(role="admins"),
            ],
        )
```

We maybe instead want to allow all org user to be able to restart builds at least (if not just everything, you cannot do much harm in the webui).

Mic92:
The react frontend that is used for the project view still has a bug regarding the login button. This is fixed in master: https://github.com/buildbot/buildbot/issues/7107

Mic92:
> > Note: For buildbot-nix, we also currently need [Mic92/nixpkgs@3bb2151](https://github.com/Mic92/nixpkgs/commit/3bb2151753ed7f1c4a29e2d5983f626b3c4a6107) because of a recent change in systemd's LoadCredential.
> 
> We discussed this on matrix, you were going to move the patch to https://github.com/Mic92/buildbot-nix?

Probably the best thing to do. Just haven't got to it.

Mic92:
The patch is now in buildbot-nix itself. The auth system can be now overwritten from the outside thanks to https://github.com/Mic92/buildbot-nix/commit/eac9acc0413a3d4e01783541e5b15135e4e666d3#diff-df8c266d76f942a320d71b583a24da5fa8ecd8135993a696f376dbd960359be7R774

flake.nix
```diff
@@ -23,6 +23,11 @@
     # actually not used when using the modules but than nothing ever will try to fetch this nixpkgs variant
     srvos.inputs.nixpkgs.follows = "nixpkgs";
 
+    buildbot-nix.url = "github:Mic92/buildbot-nix";
```
Mic92, line 26: I would maybe pin this in the beginning to a commit so that flake auto-updates don't break stuff. And then we live-test upgrades. I am still working on rolling this out in other places.

modules/nixos/buildbot-master/default.nix
```diff
@@ -0,0 +1,47 @@
...
+  sops.secrets.github-webhook-secret = buildbotSecrets;
+  sops.secrets.nix-workers = buildbotSecrets;
+
+  services.buildbot-nix.master = {
+    enable = true;
+    buildSystems = [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
+    domain = "buildbot.nix-community.org";
+    workersFile = config.sops.secrets.nix-workers.path;
+    github = {
+      tokenFile = config.sops.secrets.github-token.path;
```
zowoq, line 30: @Mic92 What permissions are needed for this token and the `nix-community-buildbot` user?

zowoq: I guess we need to make `nix-community-buildbot` an org owner otherwise we'd need manually add it as a collaborator to each repo.

Mic92: Yeah. looks like it.

zowoq:
@Mic92

Is there a reason you're using [`buildbot-worker`](https://github.com/Mic92/buildbot-nix/blob/main/nix/worker.nix#L35) instead of the default `bbworker`?

Also [this patch](https://github.com/Mic92/nixpkgs/commit/7a32f1a09082731501a49dd213b95662ab7dcc9e) changes `buildbot` and `bbworker`?

Mic92:
> @Mic92
> 
> Is there a reason you're using [`buildbot-worker`](https://github.com/Mic92/buildbot-nix/blob/main/nix/worker.nix#L35) instead of the default `bbworker`?
> 
> Also [this patch](https://github.com/Mic92/nixpkgs/commit/7a32f1a09082731501a49dd213b95662ab7dcc9e) changes `buildbot` and `bbworker`?

I didn't wanted those service accounts to show up in login manager. Also "normal" user might be granted additional privileges that I would apply selectively to buildbot.

Mic92:
But thanks for the heads-up. I forgot about this patch.

PR #896, opened 2023-10-25
secrets.yaml: add nix-community-buildbot

zowoq:


Mic92:
For buildbot-nix, we also currently need https://github.com/Mic92/nixpkgs/commit/3bb2151753ed7f1c4a29e2d5983f626b3c4a6107 because of a recent change in systemd's LoadCredential.

PR #900, opened 2023-10-27
darwin community builder

zowoq:
~~Need to delete hercules ci state, cache/cluster secrets when this is merged.~~

~~Also `/etc/ssh/authorized_keys.d/nix`.~~


Mic92:
The other repo has been moved now.

modules/darwin/community-builder/users.nix
```diff
@@ -0,0 +1,128 @@
...
+        home = "/Users/${u.name}";
+        createHome = true;
+        shell = "/bin/zsh";
+      };
+    })
+    users);
+
+  users.knownUsers = builtins.map (u: u.name) users;
+
+  users.forceRecreate = true;
```
zowoq, line 118: `forceRecreate` somehow breaks cachix deploy, I've disabled it for now so we don't need to deploy manually.

PR #921, opened 2023-11-02
modules/darwin/community-builder: add user (adisbladis)

adisbladis:


zowoq:
Do you want sudo on this box? At the moment giving admins their own accounts with sudo is a bit awkward, easier for us to share the `hetzner` account for now.

https://github.com/nix-community/infra/blob/0d98e6f5841b7d2211735682da2f049599938962/modules/darwin/common/default.nix#L15-L19

adisbladis:
> Do you want sudo on this box? At the moment giving admins their own accounts with sudo is a bit awkward, easier for us to share the hetzner account for now.

Added my key to the shared hetzner user.

modules/darwin/community-builder/users.nix
```diff
@@ -118,6 +118,11 @@ let
       trusted = true;
       uid = 546;
     }
+    {
+      name = "adisbladis";
+      trusted = true;
+      uid = 547;
+    }
```
zowoq, line 125:
```suggestion
```
Sorry if I wasn't clear, I meant for this to be removed and only use the hetzner user.

adisbladis: Oh. That's also reasonable.

PR #935, opened 2023-11-05
add cachix deploy effect

zowoq:
Switch to effects so we can have deployments that support IFD.

Functionally the darwin effect is the same as the action.

Need to deploy to build 02/03 and re-enable hercules for this repo before merging.


modules/nixos/common/auto-upgrade.nix
```diff
@@ -2,7 +2,7 @@
 {
   system.autoUpgrade.enable = lib.mkDefault true;
   system.autoUpgrade.flake = "github:nix-community/infra";
-  system.autoUpgrade.dates = "hourly";
+  system.autoUpgrade.dates = "0/2:00";
```
zowoq, line 5: I don't want to disable autoupgrade yet as I think some of our services may sort of depend on it, two hours should give us a window to see something has gone down while not being down too long that it annoys people.

PR #946, opened 2023-11-10
.github/workflows/flake-updates-input.yml: rename, add input

zowoq:
Automates creating a flake update PR for a single input.

### DIFF
--- a/.github/workflows/flake-updates-input.yml
+++ b/.github/workflows/flake-updates-input.yml
@@ -1,7 +1,12 @@
-name: "Update flakes - nixpkgs-update"
+name: "Update flakes - input"
 on:
   repository_dispatch:
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


.github/workflows/flake-updates-input.yml
```diff
@@ -1,10 +1,16 @@
-name: "Update flakes - nixpkgs-update"
+name: "Update flakes - input"
 on:
   repository_dispatch:
   workflow_dispatch:
+    inputs:
+      flakeInput:
+        description: "flake input to update"
+        required: true
+        type: string
```
zimbatm, line 9: I didn't get where this is passed from.

zowoq: There would be another field for text input next to the branch drop down.

input

PR #950, opened 2023-11-11
modules/nixos/monitoring/alert-rules: add BuildbotNoWorkers, TelegrafPluginFailing

zowoq:


modules/nixos/monitoring/alert-rules.nix
```diff
@@ -1,196 +1,42 @@
...
+        "borgbackup-job-github-org.service"
+        "borgbackup-job-nixpkgs-update.service"
+      ]
+        (name: {
+          expr = ''absent_over_time(task_last_run{name="${name}"}[1d])'';
+          annotations.description = "status of ${name} is unknown: no data for a day";
+        })) //
+      {
+        BuildbotNoWorkers = {
+          expr = ''buildbot_workers_running_total == 0'';
```
zowoq, line 15: Janitor is a worker, need to do something like < 2.

PR #955, opened 2023-11-12
modules/nixos/common: disable execWheelOnly

zowoq:


modules/nixos/common/default.nix
```diff
@@ -12,6 +12,9 @@
     inputs.srvos.nixosModules.server
   ];
 
+  # breaks smartctl in the srvos telegraf mixin
+  security.sudo.execWheelOnly = lib.mkForce false;
```
Mic92, line 16: @zimbatm makes me wonder if we should have this in srvos than.
It seems to be a common pattern to grant applications access to certain white-listed commands with sudo. Alternative would be to make it a setuid wrapper that only telegraf can execute.

Mic92:
@zowoq https://github.com/nix-community/srvos/pull/303 might be the better solution.

PR #961, opened 2023-11-14
modules/nixos/community-builder: add glepage user

GaetanLepage:
I am a regular nixpkgs contributer and would like to use the community builder to test my changes more efficiently.

I have read and understood the implication of using such shared builder as explained in the aarch64-build-box README.


Mic92:
@mergify queue

GaetanLepage:
I didn't pay attention but I just noticed that I am not marked as trusted... Should I add `trusted = true;` to my user declaration ?

zowoq:
We don't have any trusted users on this machine, see https://github.com/nix-community/infra/pull/735.

GaetanLepage:
> We don't have any trusted users on this machine, see #735.

Ok I see !
However, I don't seem to be able to build from my local machine with this remote-builder configured:
```
error: cannot add path '/nix/store/chcvbs2qfc1han25g7l9zvragvkppibi-libuv-1.47.0' because it lacks a signature by a trusted key
```

Is this an issue on my end ?

Here is the relevant line in `/etc/nix/machines`:
```
ssh://glepage@build-box.nix-community.org i686-linux,riscv64-linux,x86_64-linux /run/agenix/linux-build-box-ssh-key 1 1 benchmark,big-parallel,kvm,nixos-test - c3NoLWVkMjU1MTkgQUFBQUMzTnphQzFsWkRJMU5URTVBQUFBSUVsSVE1NHFBeTdEaDYzckJ1ZFlLZGJ6SkhycmJyck1YTFlsN1BrbWs4OEgK
```

GaetanLepage:
I also tried with `ssh-ng` instead of regular `ssh` and I get the same error.

GaetanLepage:
Actually, I think that is caused by the fact that this specific derivation was partially built on my system or somewhere else.
I tried building something else and it now works fine.
I was precisely in the "Known failure mode" / "should fail" case of this PR: https://github.com/NixOS/nix/pull/3921

zowoq:
Everyone is trusted now: https://github.com/nix-community/infra/pull/962.

PR #971, opened 2023-11-19
remove web01 and lemmy

zowoq:


adisbladis (approved):

Mic92:
Do you not just want to repurpose the machine for something else (i.e. the age key?)

zowoq:
I don't think we have any use for it currently and it is easy enough to create another if we do need one.

Mic92 (approved):

devdoc/hosts.md
```diff
@@ -48,14 +48,6 @@
...
 - Drives: 256GB SSD
 
-### `web01`
-
-- Provider: Hetzner
-- Instance type: [CX31](https://www.hetzner.com/cloud)
-- CPU: 2 vCPUs on Intel Xeon
-- RAM: 8GB
-- Drives: 80GB SSD
-
```
Mic92, line 58: Maybe keep this info here, so we have a full list of potential hardware to use.

zowoq: This is a list of machines that we are currently using, I don't understand why we would keep it here?

Mic92: Because not everyone has insights in what hardware we have available. I don't see the advantage in removing it either, but don't consider this as a reason to block this pr.

PR #983, opened 2023-11-30
build02/nixpkgs-update: add another worker

zowoq:
Haven't tried this for a while, memory needed for evals may still be an issue.

With more projects using buildbot there is less need for another hercules agent. https://github.com/nix-community/infra/pull/990

Each nixpkgs-update worker is throttled to 1 PR an hour so adding another worker won't really make the ofborg situation much worse than it is currently but it will allow nixpkgs-update to process the updates much faster.

### DIFF
--- a/hosts/build02/nixpkgs-update.nix
+++ b/hosts/build02/nixpkgs-update.nix
@@ -191,8 +191,8 @@ in
 
   systemd.services.nixpkgs-update-worker1 = mkWorker "worker1";
   systemd.services.nixpkgs-update-worker2 = mkWorker "worker2";
+  systemd.services.nixpkgs-update-worker3 = mkWorker "worker3";
   # Too many workers cause out-of-memory.
-  #systemd.services.nixpkgs-update-worker3 = mkWorker "worker3";
   #systemd.services.nixpkgs-update-worker4 = mkWorker "worker4";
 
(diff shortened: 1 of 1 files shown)


zowoq:
cc @Mic92 @ryantm @rhendric

rhendric:
When I tested most recently, the command nixpkgs-update uses to compare outpaths eats something like 24 GiB. Does build02 have enough memory to support three of those running concurrently? If so, :+1:.

An alternative I've been considering to improve nixpkgs-update throughput despite OfBorg throttling would be to separate out just the PR submission into its own queue and service, so that workers don't have to sit idle while the throttle time elapses (unless the queue fills to some fixed capacity, I guess).

zowoq:
```
[zowoq@build02:~]$ free
               total        used        free      shared  buff/cache   available
Mem:        65786960    45979608    17469440         920     2337912    19217664
Swap:       32893436      228968    32664468
```

Might be fine if it is just the nixpkgs-update workers and no hercules.

> An alternative I've been considering to improve nixpkgs-update throughput despite OfBorg throttling would be to separate out just the PR submission into its own queue and service, so that workers don't have to sit idle while the throttle time elapses (unless the queue fills to some fixed capacity, I guess).

I've been thinking of something like this as well, separate queues for PR submission and evals would be good.

PR #1000, opened 2023-12-04
darwin/community-builder: add kranzes

Kranzes:
I'd like to be able to use the darwin box for quickly testing out nixpkgs PRs without waiting ages for ofborg.


Kranzes:
Thanks! <3

PR #1011, opened 2023-12-07
build02: remove cachix secret key

zowoq:


Mic92:
Has the ryantm bot its own cachix key?

Mic92:
Ah. This is only rotating they key, not removing.

zowoq:
This secret has a token and a signing key, this just removes the signing key.

Mic92 (approved):

PR #1012, opened 2023-12-07
modules/shared/builder: add gc.gbFree option

zowoq:


modules/shared/builder.nix
```diff
@@ -1,6 +1,11 @@
...
 {
-  nix.gc.options = ''
-    --max-freed "$((50 * 1024**3 - 1024 * $(df -P -k /nix/store | tail -n 1 | ${pkgs.gawk}/bin/awk '{ print $4 }')))"
+  options.nixCommunity.gc.min = lib.mkOption {
+    type = lib.types.int;
+    default = 50;
+  };
+
+  config.nix.gc.options = ''
+    --max-freed "$((${toString config.nixCommunity.gc.min} * 1024**3 - 1024 * $(df -P -k /nix/store | tail -n 1 | ${pkgs.gawk}/bin/awk '{ print $4 }')))"
```
Mic92, line 9: Maybe some comment what this line does?

modules/shared/builder.nix
```diff
@@ -1,6 +1,11 @@
-{ pkgs, ... }:
+{ config, lib, pkgs, ... }:
 {
-  nix.gc.options = ''
-    --max-freed "$((50 * 1024**3 - 1024 * $(df -P -k /nix/store | tail -n 1 | ${pkgs.gawk}/bin/awk '{ print $4 }')))"
+  options.nixCommunity.gc.min = lib.mkOption {
+    type = lib.types.int;
+    default = 50;
```
Mic92, line 5: Is this percent?

zowoq: GB.

modules/shared/builder.nix
zowoq: I'll leave this for now.

Mic92: the module description makes it better.

PR #1023, opened 2023-12-11
unpin terraform shell

zowoq:
We likely won't return to running terraform automatically on CI for a while, if ever.

Don't really need the pin for manual use so we may as well save ourselves an extra input.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -238,7 +238,6 @@
         "nur-update": "nur-update",
         "sops-nix": "sops-nix",
         "srvos": "srvos",
-        "tf-pkgs": "tf-pkgs",
         "treefmt-nix": "treefmt-nix"
       }
     },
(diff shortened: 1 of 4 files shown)


Mic92 (approved):

zowoq:
> Are you saying this because of the move to buildbot-nix?

No, automated terraform may be nice but for the limited changes we have I don't think it has been worth the effort that I'd put into it.

> I think it should still be possible to trigger GitHub Actions after buildbot has returned its checks.

I've already looked at this, buildbot only sets a check status (currently 34 of them), not check run or check suite so it is rather awkward to use. Could improve the checks it but I'd rather have support for running deployment type stuff in effects or something similar on buildbot so we can unify on a single CI platform instead of bridging them.

zowoq:
I had someone suggest to me that using https://github.com/suzuki-shunsuke/tfcmt, github actions and chatops comments (`/plan` `/apply`) might be a compromise for this type of scenario, e.g. not integrated into CI / blocking but also not needing to drop into a shell to apply changes.

As an example I ran plan manually on https://github.com/nix-community/infra/pull/1054#issuecomment-1872408658.

I'd likely want to wait until after we have dealt with effects on buildbot before following up on this but thought I'd mention it now anyway.

PR #1024, opened 2023-12-11
modules/darwin/common/flake-inputs: fix inputs

zowoq:
This was causing unnecessary rebuilds.

### DIFF
--- a/modules/darwin/common/flake-inputs.nix
+++ b/modules/darwin/common/flake-inputs.nix
@@ -3,7 +3,7 @@
 {
   services.telegraf.extraConfig.inputs.file =
     let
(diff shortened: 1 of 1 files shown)


modules/darwin/common/flake-inputs.nix
```diff
@@ -3,7 +3,7 @@
 {
   services.telegraf.extraConfig.inputs.file =
     let
-      inputsWithDate = lib.filterAttrs (_: input: input ? lastModified) inputs;
+      inputsWithDate = lib.filterAttrs (_: input: input ? lastModified) inputs.self.inputs;
```
Mic92, line 6: Does this include basically the flake itself?

zowoq: Yes, it was including the flake.

PR #1028, opened 2023-12-12
darwin/community-builder: add user sternenseemann

sternenseemann:
Please there is an angry mob of aarch64-darwin users in front of my house that want me to fix aarch64-darwin Haskell issues.


sternenseemann:
Thanks :)

PR #1032, opened 2023-12-13
modules/nixos/buildbot: disable `buildbotNetUsageData`

zowoq:


Mic92:
The information used doesn't seem excessive: http://docs.buildbot.net/current/manual/configuration/global.html#buildbotnetusagedata
I'll don't have an opinion if it should be enabled or not. I'll keep it on for buildbot-nix and leave the decision to its users.

zimbatm:
Having telemetry enabled could be positive for us in this instance. All the data is already public. What this does is bring awareness of our usage to the main project, with the associated potential upsides.

PR #1033, opened 2023-12-13
build02/nixpkgs-update: add more workers

zowoq:
Will wait a few days before trying this but I think it should be fine, with the new machine the constraint will likely be if ofborg can keep up with the volume.


zowoq:
Deployed this manually, will let it run for a bit to see if ofborg will keep up.

PR #1037, opened 2023-12-16
modules/nixos/common/security: drop fail2ban

zowoq:
We only allow publickey auth so this is only of limited value anyway.

It has been disabled for the last couple of months as the nixpkgs module was broken and I forgot to re-enable it.


zimbatm (approved):
💯 in my experience, the only person I prevented getting access to the server was myself.

Mic92 (approved):

PR #1038, opened 2023-12-16
 modules/nixos/monitoring/alert-rules: add NixpkgsOutOfDate2Weeks

zowoq:
Bumping the NixpkgsOutOfDate alert to two weeks so we can switch to nixos-unstable.


zowoq:
With the new aarch64-linux machine it is only darwin that may occasionally cause problems with excess rebuilds. Okay for now but might need to do this if it becomes frequent.

PR #1044, opened 2023-12-19
build01,build02,build03: remove hostId

zowoq:
This was needed for zfs which we aren't using on these hosts.


Mic92:
We could even have a single hostId in the zfs profile as this is otherwise irrelevant for anyone not using network-based block devices. But now we only have one machine left so probably not worth the trouble.

PR #1058, opened 2023-12-27
switch to effect for pages

zowoq:


zowoq:
If we don't end up needing effects for deploying the hosts (https://github.com/nix-community/infra/pull/1059) I don't see much benefit in using effects just for pages.

@Mic92 As we don't have a strong use case for buildbot-nix effects here at the moment not sure if this will alter you plans about implementing them in buildbot-nix.

Mic92:
It's still on my roadmap to finish the implementation of them

zowoq:
I'll merge just the effect in this PR to make sure it works and I'll do follow up for removing the action and updating terraform. https://github.com/nix-community/infra/pull/2304

PR #1059, opened 2023-12-27
switch to effect for darwin ssh deploy

zowoq:


zowoq:
Now that we're in the process to switching to comin for nixos, adding support to comin for nix-darwin and having consistent deployments makes sense, closing.

dev/effect.nix
```diff
@@ -0,0 +1,35 @@
...
+          {
+            darwin02 = hci-effects.runNixDarwin {
+              ssh.destination = "hetzner@darwin02.nix-community.org";
+              configuration = self.darwinConfigurations.darwin02;
+              buildOnDestination = true;
+              secretsMap.ssh-deployment = "ssh-deployment";
+              userSetupScript = ''
+                writeSSHKey ssh-deployment
+                cat >>~/.ssh/known_hosts <<EOF
+                darwin02.nix-community.org ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAICJqwpMUEl1/iwrBakeDb1rlheXlE5mfDLICVz8w6yi6
```
Mic92, line 16: I probably should test examples like that in my deployments as well.

zowoq:
https://buildbot.nix-community.org/#/builders/3618/builds/23

```diff
diff --git a/buildbot_effects/buildbot_effects/__init__.py b/buildbot_effects/buildbot_effects/__init__.py
index 7b2273e90..155eed6be 100644
--- a/buildbot_effects/buildbot_effects/__init__.py
+++ b/buildbot_effects/buildbot_effects/__init__.py
@@ -190,6 +190,8 @@ def run_effects(
     env["HERCULES_CI_SECRETS_JSON"] = "/run/secrets.json"
     env["NIX_BUILD_TOP"] = "/build"
     env["TMPDIR"] = "/tmp"  # noqa: S108
+    env["NIX_REMOTE"] = "daemon"
+    env["NIX_STORE"] = "/nix/store"
     clear_env = set()
     clear_env.add("TMP")
     clear_env.add("TEMP")
@@ -235,6 +237,9 @@ def run_effects(
         "/nix/store",
         "--hostname",
         "hercules-ci",
+        "--bind",
+        "/nix/var/nix/daemon-socket/socket",
+        "/nix/var/nix/daemon-socket/socket",
     ]
 
     with NamedTemporaryFile() as tmp:
```

.github/workflows/darwin.yml
```diff
@@ -1,60 +0,0 @@
...
-          echo -e "$known_hosts" >>"$HOME/.ssh/known_hosts"
-          cat >>"$HOME/.ssh/config" <<EOF
-          Host *
-            IdentityFile ~/.ssh/id_ed25519
-            IdentitiesOnly yes
-          EOF
-      - name: Deploy
-        run: |
-          ./inv deploy --hosts ${{ matrix.host }}
-      - name: Alert on failure
```
zowoq, line 49: We won't have this alert with effects but if it fails the buildbot status check will be red on master and darwin deploy failures were mostly a just problem on the m1s from hetzner, it has only failed a couple of times since we switched to the better m2/m4 machines.

zowoq:
@Mic92 effect failures don't seem to be included in the github status checks.

https://buildbot.nix-community.org/#/builders/6/builds/1260 fails but nix-eval, nix-build are both green.

Mic92:
Cc @magicrb

MagicRB:
Will look into today

PR #1068, opened 2024-01-04
web01: init

zowoq:


hosts/web01/hardware-configuration.nix
```diff
@@ -0,0 +1,50 @@
...
+
+  disko.devices = {
+    disk = {
+      sda = {
+        type = "disk";
+        device = "/dev/sda";
+        content = {
+          type = "gpt";
+          partitions = {
+            ESP = {
```
zowoq, line 25: Not sure if bootable, fs-type are still needed?

https://github.com/nix-community/infra/commit/8713cd3c58c4cac32372f10ffdcbec7a5a350630
```
           {
              name = "ESP";
              start = "0";
              end = "1GiB";
              fs-type = "fat32";
              bootable = true;
              content = {
                type = "filesystem";
                format = "vfat";
                mountpoint = "/boot";
              };
```

Mic92: `fs-type` is only available for the legacy table type. With the new GPT type there are GPT codes like `type = "EF00";`

zowoq:
We haven't used or needed this machine in the last three months and as it is on an external oracle account which is a bit less convenient for the other admins to access. I think if we need another vm for hosting experiments or a fallback builder for aarch64 we can spin a hetzner vm easily enough.

PR #1070, opened 2024-01-07
modules/nixos/common: use latest kernel

zowoq:
https://github.com/nix-community/infra/commit/f3967527610aebf42793a6553a73eb8ad5601876

Now we're not using zfs we can go back to using the latest kernel.


zimbatm:
Are there any specific features that we need from the latest kernel?

Mic92:
Not all security fixes always gets backported in the Linux kernel. And our CI security depends on the Linux kernel to not have bugs.

zimbatm (approved):

PR #1071, opened 2024-01-07
use ssh-ng by default

zowoq:


zimbatm:
The last time I checked (that was a while ago), ssh-ng didn't implement all the features of the ssh protocol. Is there any concrete benefit to switching?

Mic92:
It allows remote builds with an untrusted user? What features missing for ssh-ng? I am using this for ages now on my own system but haven't encountered any limitations.

zowoq:
> ssh-ng didn't implement all the features of the ssh protocol

I think it is still missing support for some cli options but that isn't an issue for remote builders in `/etc/nix/machines`.

> Is there any concrete benefit to switching?

TBH I had assumed ssh-ng was preferred, e.g. https://github.com/nix-community/srvos/pull/58.

zimbatm:
I think we all assume that because it's "NG", it's better. But I forgot the details and it was 2 years ago, so things might have changed. I guess the only way to find out is to try it.

PR #1073, opened 2024-01-07
remote-workers: enable kvm/nixos-test support on build04

Mic92:


modules/nixos/remote-workers.nix
```diff
@@ -12,6 +12,11 @@
       sshKey = config.sops.secrets.id_buildfarm.path;
       sshUser = "nix";
       system = "aarch64-linux";
+      supportedFeatures = [
```
zowoq, line 15: Should we add `﻿gccarch-armv8-a` here as well?

PR #1076, opened 2024-01-09
darwin/community-builder: add user corngood

corngood:
Hi, I'd like to test building dotnet sdks from source (https://github.com/NixOS/nixpkgs/pull/190129) on darwin.  Thanks!


zowoq:
We don't really have specific rules for allowing access to this machine but it is basically limited to people who are well known in the nixpkgs/nixos/etc community, is there somebody who is known (e.g. nixpkgs committer) who can vouch for you?

corngood:
I was linked here by @raphaelr, but it's a lot to ask for a recommendation.

I have 152 commits in nixpkgs since 2016, and 6 in nix since 2017, according to shortlog.

Mic92:
I remember your contributions. Looks good to me.

zowoq:
@corngood This machine only has a 256GB disk, unfortunately the size of dotnet builds means it fills very quickly. Yesterday I removed the `keep-failed` build outputs, I'll probably need to do the same every day or two to keep the machine usable for others.

corngood:
@zowoq Thanks, and apologies for the inconvenience.  As far as I could tell I had no way to delete those builds myself.  I'll try to avoid `--keep-failed` unless absolutely necessary.  I think I've solved most of the problems on aarch64-darwin, so I'm mostly just doing cleanup now.  I'm running some tests on x64 as well, but it seems to be going extremely slowly.  If you need to kill anything that's cool.  Also feel free to email or message me on matrix @(github username):(github username).com.

corngood:
@zowoq I feel like I'm fighting the nix garbage collector, and /nix seems to be pretty full.

I was trying to bisect something, which requires pulling pretty large closures.  Should I make a point of running GC myself?

zowoq:
The machine runs gc every hour but it won't remove anything unless there is less than 30GB free space on disk.

I don't think anyone will mind if you gc the store every day or two.

With the build outputs in /tmp, users with GBs in $HOME and the space macos itself occupies there really just isn't enough space for massive closures on a shared machine. We are looking at options for increasing storage, hopefully we can do something about it in the next few weeks.

PR #1088, opened 2024-01-15
README, .github/settings.yml: misc updates

zowoq:


.github/settings.yml
```diff
@@ -5,13 +5,13 @@ repository:
   name: infra
 
   # A short description of the repository that will show up on GitHub
-  description: nix-community infrastructure [maintainer=@Mic92]
```
zowoq, line 8: I'd put `@nix-community/admin` here but that is only visible to members of `@nix-community`.

zimbatm: then let's put you, so there is at least one point of contact

zowoq: Okay, updated.

PR #1095, opened 2024-01-21
modules/*/community-builder: add user teto

teto:


zowoq:
Need to add your key at `modules/darwin/community-builder/keys/teto` for the darwin user as well.

teto:
ty. I was not sure if I could symlink modules/darwin/community-builder/keys/teto to modules/nixos/community-builder/keys/teto so I duplicated the key

PR #1100, opened 2024-01-25
refactor supportedFeatures

zowoq:


modules/nixos/remote-workers.nix
```diff
@@ -10,13 +10,8 @@
...
       sshUser = "nix";
-      system = "aarch64-linux";
-      supportedFeatures = [
-        "big-parallel"
-        "kvm"
-        "nixos-test"
-        "gccarch-armv8-a"
-      ];
+      systems = [ "aarch64-linux" ];
+      supportedFeatures = inputs.self.outputs.nixosConfigurations.build04.config.nix.settings.system-features;
```
zowoq, line 14: I can't find the PR but we discussed doing this a while ago and decided against it to save a bit of eval performance. Since then we've merged https://github.com/nix-community/infra/pull/669 so I think we can to merge this now.

PR #1104, opened 2024-01-29
build02/nixpkgs-update: remove logs older than 18 months

zowoq:
Mostly just want to remove cruft, e.g. deleted packages, outdated package sets. I think 18 months is a reasonable window?

cc @rhendric 

Copied service from:
https://github.com/nix-community/infra/blob/6a302a08cb0ffae537caacaa85f89fea9d340bbf/modules/nixos/hydra.nix#L15-L19

### DIFF
--- a/hosts/build02/nixpkgs-update.nix
+++ b/hosts/build02/nixpkgs-update.nix
@@ -231,6 +231,15 @@ in
     '';
   };
 
+  systemd.services.nixpkgs-update-delete-old-logs = {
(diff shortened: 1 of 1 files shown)


rhendric (approved):

zowoq:
I don't think we need to keep the nixpkgs-update worker logs for 18 months so I've opened another PR to remove them after 3 months. https://github.com/nix-community/infra/pull/1109

PR #1106, opened 2024-01-30
build02/nixpkgs-update: reduce timeout to six hours

zowoq:
A few times I'll restarted workers that have been stuck for 3+ hours, not much point it letting it run until it hits the 12 hour timeout.

https://github.com/nix-community/infra/pull/951

6 hours seems to cover almost everything, after a couple of weeks we can review the logs, fix/disable problem packages or increase the timeout if necessary?

cc @rhendric

### DIFF
--- a/hosts/build02/nixpkgs-update.nix
+++ b/hosts/build02/nixpkgs-update.nix
@@ -60,7 +60,7 @@ let
       function run-nixpkgs-update {
         exit_code=0
         set -x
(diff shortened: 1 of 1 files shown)


rhendric (approved):

zowoq:
So far there has only three that seem to be a problem. https://github.com/ryantm/nixpkgs-update/pull/400

PR #1110, opened 2024-02-04
modules/nixos/disko-raid: add option for filesystem type

zowoq:
Will switch to xfs on build02 to address the inode problems, would seem to be the simplest option.

- don't really want to go back to zfs and being limited to which kernels we can use. (we also seem to have had high memory usage with zfs.)
- bcachefs is a bit too new. (also disko doesn't have support for bcache multi-disk yet).
- btrfs might be okay but disko doesn't have support for multi-disk yet.

### DIFF
--- a/modules/nixos/disko-raid.nix
+++ b/modules/nixos/disko-raid.nix
@@ -36,6 +36,11 @@ in
     inputs.disko.nixosModules.disko
   ];
   options = {
+    nixCommunity.disko.fsType = lib.mkOption {
+      type = lib.types.enum [ "btrfs" "ext4" ];
+      default = "ext4";
+      description = "Type of filesystem to use";
+    };
     nixCommunity.disko.raidLevel = lib.mkOption {
       type = lib.types.int;
       default = 1;
@@ -65,7 +70,7 @@ in
         level = config.nixCommunity.disko.raidLevel;
         content = {
           type = "filesystem";
-          format = "ext4";
+          format = config.nixCommunity.disko.fsType;
           mountpoint = "/";
         };
       };


Mic92:
xfs will zero out not synced files after a crash btw. I had reproducible data corruptions with it for git repositories on Hetzner machines - Upstream won't fix this because they don't consider unreliable hardware something where one would use XFS.

Mic92:
The original sources are gone but it's the behavior described here: https://github.com/pedmon/xfs_recover

zowoq:
> I had reproducible data corruptions with it

Which kernel version(s) was this with?

Mic92:
Last year. Likely the stock NixOS kernel at the time.

zowoq:
Shall we do something non-standard like btrfs over mdraid or shall I just not use disko on this machine and do btrfs raid0 manually?

zimbatm:
How many inodes do we need?

By default, ext4 sets the number of inodes proportional to the disk size, but it can be changed during formatting. Ext4 can support ~ 4 billion inodes ( 2^32-1 ). All you need to do is pass the new number to the disk config:
```
extraArgs = ["-N", (toString 4294967295)];
```

zowoq:
Not sure I understand the purpose of you comment? 

I said on matrix that want to switch to something that has dynamic allocation, you acked that with a 👍. 

Do you now want to stay on ext4?

zimbatm:
What I mean is that bumping the inodes on ext4 is **also** an option. If you want to switch to btrfs it's fine by me.

zowoq:
I'll try to get build02 redeploying in the coming week, I've been putting it off to ensure that I have plenty of time available in case of any problems.

zowoq:
Will finish this in https://github.com/nix-community/infra/pull/1158.

zowoq:
Besides the inode issue also seeing a lot of messages like this during builds, seems to be caused by ext4 + optimise store.

```
cannot link '/nix/store/.links/0dwi8xhyl6ykk4xbbs3iyy89gspbipc437cx055j2kf6xcwx6gvg' to '/nix/store/14grigrj2nci6fadzyq8bn4r1m399bnm-source/pkgs/tools/virtualization/distrobuilder/default.nix': No space left on device
```

As an interim fix I'm thinking of going back to zfs, at least it seemed to work. https://github.com/nix-community/infra/pull/1199

I'll look at adding multidisk bcachefs support to disko and maybe we can try it on the community builder first.

zowoq:
@Mic92 We're also getting the same `cannot link` spam on build03/build04, switching everything back to zfs seems like the easiest and safest option at the moment, are you okay with that?

Mic92:
Sure. Zfs is a safe bet just now. I am also evaluating bcachefs in personal non-critical setups. We haven't implemented multi-disk bcachefs support so far in disko.

zowoq:
I'll also investigate problem(s) I encountered with mirrored boot and see if I fix it in disko.

zowoq:
build01, build02, build04 are on zfs.

Tried build03 and it errored on boot with a zfs problem, I've reinstalled it with the original ext4 config for now.

@Mic92 IIRC you mentioned a problem with zfs on one machine, was it this?

zfs

Mic92:
Mhm. Don't remember. You could also later use the emergency shell in systemd initrd to debug this in a shell. Most of these issues are related to either missing block device driver or the zpool not being exported before reboot.

PR #1113, opened 2024-02-06
build02/nixpkgs-update: run fetchers every 12 hours

zowoq:
Spread out new github and repology PRs a little instead of batching them once a day.

I think twice a day should be fine but probably need to check API ratelimits to increase it any further.

cc @rhendric

### DIFF
--- a/hosts/build02/nixpkgs-update.nix
+++ b/hosts/build02/nixpkgs-update.nix
@@ -132,13 +132,14 @@ let
     script = ''
       mkdir -p "$LOGS_DIRECTORY/~fetchers"
       cd "$LOGS_DIRECTORY/~fetchers"
+      sleep 60 # wait for network
       while true; do
         run_name="${name}.$(date +%s).txt"
         rm -f ${name}.*.txt.part
         ${cmd} > "$run_name.part"
         rm -f ${name}.*.txt
         mv "$run_name.part" "$run_name"
-        sleep 24h
+        sleep 12h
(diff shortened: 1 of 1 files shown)


hosts/build02/nixpkgs-update.nix
```diff
@@ -132,13 +132,14 @@ let
...
       cd "$LOGS_DIRECTORY/~fetchers"
+      sleep 60 # wait for network
       while true; do
         run_name="${name}.$(date +%s).txt"
         rm -f ${name}.*.txt.part
         ${cmd} > "$run_name.part"
         rm -f ${name}.*.txt
         mv "$run_name.part" "$run_name"
-        sleep 24h
+        sleep 12h
```
zowoq, line 142: I've only seen it happen a couple of times but if the network isn't accessible the fetchers seem to get stuck, adding a sleep seems like a simple fix.

hosts/build02/nixpkgs-update.nix
```diff
@@ -132,13 +132,14 @@ let
     script = ''
       mkdir -p "$LOGS_DIRECTORY/~fetchers"
       cd "$LOGS_DIRECTORY/~fetchers"
+      sleep 60 # wait for network
```
zowoq, line 135: I've only seen it happen a couple of times but if the network isn't accessible the fetchers seem to get stuck, adding a sleep seems like a simple fix.

PR #1132, opened 2024-03-01
darwin: disable netbios on activation

zimbatm:
We have received a notification from the German Federal Office for Information Security (BSI) about our NetBIOS being enabled, and it potentially being used for DDoS reflection attacks.

### DIFF
--- a/modules/darwin/common/default.nix
+++ b/modules/darwin/common/default.nix
@@ -66,5 +66,12 @@
     StrictModes no
   '';
 
+  # Make sure to disable netbios on activation
+  system.activationScripts.postActivation.text = ''
+    echo disabling netbios... >&2
+    launchctl disable system/netbiosd
+    launchctl unload -w /System/Library/LaunchDaemons/com.apple.netbiosd.plist 2>/dev/null || true
+  '';
+
   time.timeZone = "GMT";
 }


zimbatm:
I already disabled the service manually, so this needs to be tested.

modules/darwin/common/default.nix
```diff
@@ -66,5 +66,11 @@
     StrictModes no
   '';
 
+  # Make sure to disable netbios on activation
+  system.activationScripts.no-netbios.text = ''
```
zowoq, line 70: `nix-darwin` doesn't have support for arbitrary activationScripts, need to use pre/post.
```suggestion
  system.activationScripts.postActivation.text = ''
```

zimbatm: Good to know, thanks

zowoq:
> so this needs to be tested.

Have you done this yourself or do you want me to do it?

zimbatm:
thanks for offering. I deployed the change on darwin02 and darwin03 and added a log line to make the UX look nicer.

For the deploy to work, I also had to patch the tasks.py:
```diff
--- a/tasks.py
+++ b/tasks.py
@@ -27,7 +27,7 @@ def deploy_nixos(hosts: List[DeployHost]) -> None:
             command = "sudo nixos-rebuild"
 
         res = h.run_local(
-            ["nix", "flake", "archive", "--to", f"ssh://{h.host}", "--json"],
+            ["nix", "flake", "archive", "--to", f"ssh://hetzner@{h.host}", "--json"],
             stdout=subprocess.PIPE,
         )
         data = json.loads(res.stdout)
```

zowoq:
> For the deploy to work, I also had to patch the tasks.py

Looks like I dropped the user when we switched from `rsync` to `nix flake archive`, I have `User hetzner` set in my ssh config (and also in the github deploy action) so never noticed that it wasn't working. https://github.com/nix-community/infra/pull/1136

PR #1162, opened 2024-03-18
build02/nixpkgs-update: reuse worker nixpkgs repo for updatescript

zowoq:
I've tested this a couple of times and it seems fine.


hosts/build02/packages-with-update-script.nix
```diff
@@ -1,5 +1,5 @@
 let
-  pkgs = import /var/cache/nixpkgs-update/updatescript/nixpkgs { };
+  pkgs = import /var/cache/nixpkgs-update/worker/nixpkgs { };
```
zowoq, line 2: @ryantm Was there a reason you used a separate repo for this?

PR #1164, opened 2024-03-21
Revert "build02: switch to btrfs"

zowoq:
This reverts commit 487138275cd3c02acc6847ee3c9d7cd2544400fa.

### DIFF
--- a/hosts/build02/configuration.nix
+++ b/hosts/build02/configuration.nix
@@ -11,7 +11,6 @@
     inputs.self.nixosModules.disko-raid
   ];
 
-  nixCommunity.disko.fsType = "btrfs";
   nixCommunity.disko.raidLevel = 0; # more disk space, we don't have much state to restore anyway
 
   networking.hostName = "build02";


Mic92:
Did btrfs cause issues?

zowoq:
Yeah, enough that it was more hassle than what we were using originally.

zowoq:
@Mic92 I forgot to mention that a recent change in disko seems to have broken the mdraid config that we are using here, https://github.com/nix-community/disko/commit/fe064a639319ed61cdf12b8f6eded9523abcc498 was the good commit that I ended up using for this reinstall. I didn't save the error as I assumed that I could reproduce it with the vm tests but they seem to fail with a different error than the one I saw on both good and bad commits.

Mic92:
Is systemd in initrd enabled?

zowoq:
No, it isn't enabled.

Mic92:
Ok. I am not using mdraid very often, but I am also not aware of regressions at the moment.

PR #1166, opened 2024-03-22
comin: add telegraf and alert rule

zowoq:


modules/nixos/monitoring/alert-rules.nix
```diff
@@ -11,6 +11,12 @@
           annotations.description = "status of ${name} is unknown: no data for a day";
         })) //
       {
+        CominDeploymentFailing = {
+          expr = ''comin_deployment_status != 2'';
+          for = "30m";
```
zowoq, line 16: web02 is a bit slow when switching so will leave this a bit longer then necessary for now.

PR #1167, opened 2024-03-22
modules/nixos/monitoring: remove grafana

zowoq:
Added this back in september and didn't end up doing anything with it. We seem to have be doing okay without it, can always add it back we we do find a need for it.


zimbatm (approved):
Interesting take. I always assumed that Grafana was essential in a monitoring stack. But maybe not. Let's try.

PR #1183, opened 2024-03-31
modules/nixos/common/comin: add prometheus, build04: switch to comin

zowoq:
Will try this for another week or so on two hosts with the prometheus metrics and then look at dropping autoupgrade.

### DIFF
--- a/modules/nixos/common/comin.nix
+++ b/modules/nixos/common/comin.nix
@@ -4,6 +4,10 @@
     inputs.comin.nixosModules.comin
   ];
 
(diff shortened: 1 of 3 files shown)


modules/nixos/common/comin.nix
```diff
@@ -4,6 +4,10 @@
     inputs.comin.nixosModules.comin
   ];
 
+  services.telegraf.extraConfig.inputs.prometheus.urls = [
+    "http://localhost:4243/metrics"
```
zowoq, line 8: I'll do a follow up PR for the alert rules.

PR #1199, opened 2024-04-06
build01: switch to zfs

zowoq:
Had issues with mirrored boot so not going to bother with it for now.

@Mic92 This config passes the nixos-anywhere vm-test but can you take a look to make sure the zfs stuff is okay?

### DIFF
--- a/modules/nixos/common/default.nix
+++ b/modules/nixos/common/default.nix
@@ -1,4 +1,4 @@
-{ inputs, pkgs, ... }:
+{ config, inputs, pkgs, ... }:
 {
   imports = [
     ./auto-upgrade.nix
@@ -29,7 +29,7 @@
 
   srvos.flake = inputs.self;
 
-  boot.kernelPackages = pkgs.linuxPackages_latest;
+  boot.kernelPackages = pkgs.lib.mkIf (!config.boot.supportedFilesystems.zfs or false) pkgs.linuxPackages_latest;
 
   zramSwap.enable = true;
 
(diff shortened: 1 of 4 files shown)


modules/nixos/disko-zfs.nix
```diff
@@ -0,0 +1,89 @@
...
+    zpool = {
+      zroot = {
+        type = "zpool";
+        options = {
+          ashift = "12";
+        };
+        rootFsOptions = {
+          acltype = "posixacl";
+          atime = "off";
+          compression = "off";
```
Mic92, line 74: Wouldn't lz4 makes sense here?

```suggestion
          compression = "lz4";
```

modules/nixos/common/default.nix
```diff
@@ -29,7 +29,7 @@
 
   srvos.flake = inputs.self;
 
-  boot.kernelPackages = pkgs.linuxPackages_latest;
+  boot.kernelPackages = pkgs.lib.mkIf (!config.boot.supportedFilesystems.zfs or false) pkgs.linuxPackages_latest;
```
Mic92, line 32:
```suggestion
  boot.kernelPackages = if config.boot.supportedFilesystems.zfs or false then pkgs.zfsUnstable.latestCompatibleLinuxPackages else pkgs.linuxPackages_latest;
```

zowoq: Following the stable kernel is intentional, don't want kernel versions downgrades, e.g. in our next nixpkgs bump it'll go from 6.7 -> 6.6.

zowoq:
I wasn't planning to put zfs on build01 but seemed like a good opportunity to test this config.

PR #1207, opened 2024-04-08
build03: switch to zfs

zowoq:


Mic92:
I have actually two other machines with zfs on this hetzner machine model. I can compare later.

zowoq:
> I have actually two other machines with zfs on this hetzner machine model. I can compare later.

@Mic92 Any ideas what the issue is which this machine?

zowoq:
In the next couple of weeks I'm going to try https://github.com/nix-community/infra/pull/1252 on build01 and then on build03.

If I still have zfs problems with build03 I'll just reinstall it with xfs in raid0, I want it off ext4 and I'm not really concerned about the potential problems with xfs.

PR #1221, opened 2024-04-14
build02: set zfs_arc_max

zowoq:
By default it can use up to 64GB (50%) so limiting it to 16GB seems like it might help?

___

zfs


Mic92:
Usually also having a swap partition (not zswap but the Linux one) is recommended on systems where RAM is often maxed out.
This gives zfs some headroom to give memory back to the system.

PR #1229, opened 2024-04-18
nixpkgs-update supervisor

zowoq:
I looked at packaging/wrapping `supervisor.py` but seems to make it more complicated than is really necessary.

@rhendric Does this look okay?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -87,6 +87,7 @@
               };
 
             packages = pkgs.lib.optionalAttrs defaultPlatform {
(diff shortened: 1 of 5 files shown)


rhendric:
Looks plausible? Love having the tests run CI; if they pass I have no objections. I tried looking into the most recent build result corresponding to the new test job and it looks like they were skipped—but maybe that's okay because a previous build ran them and the infrastructure is good at not re-running if no relevant inputs changed? I don't know much about flakes or the new fileset API or the buildbot infrastructure so I'm really out at sea here.

zowoq:
Yes, if it has been built and hasn't changed it'll be skipped.

PR #1235, opened 2024-04-22
modules/nixos/buildbot: set Restart, RestartSec

zowoq:
Buildbot fails if it can't connect to github after the host is rebooted.

alternative to 83f3142fd8d08169d0fab4d4a4f83bc603287371

### DIFF
--- a/modules/nixos/buildbot.nix
+++ b/modules/nixos/buildbot.nix
@@ -14,6 +14,15 @@
     "http://localhost:8011/metrics"
   ];
 
+  systemd.services.buildbot-master.serviceConfig = {
+    Restart = "on-failure";
+    RestartSec = "30s";
+  };
+  systemd.services.buildbot-worker.serviceConfig = {
+    Restart = "on-failure";
+    RestartSec = "30s";
+  };
+
   sops.secrets.buildbot-github-oauth-secret = { };
   sops.secrets.buildbot-github-token = { };
   sops.secrets.buildbot-github-webhook-secret = { };


modules/nixos/buildbot.nix
```diff
@@ -14,6 +14,15 @@
     "http://localhost:8011/metrics"
   ];
 
+  systemd.services.buildbot-master.serviceConfig = {
+    Restart = "on-failure";
```
Mic92, line 18: I can probably upstream this.

modules/nixos/buildbot.nix
```diff
@@ -14,6 +14,15 @@
...
+    RestartSec = "30s";
+  };
+  systemd.services.buildbot-worker.serviceConfig = {
+    Restart = "on-failure";
```
Mic92, line 22: Afaik the worker shouldn't be affected by this. The github fetching logic should be only loaded by the buildbot master process.

modules/nixos/buildbot.nix
zowoq: Thanks.

modules/nixos/buildbot.nix
zowoq: Does the worker have any sort of timeout if it can't connect to the master or will it idle indefinitely? I have seen it fail but can't remember the specific error.

zowoq: I'll drop it for the worker, easy enough to add it back if necessary.

https://github.com/nix-community/infra/pull/1248

modules/nixos/buildbot.nix
zowoq: Seems this needs to be always instead of on-failure as the service exits successfully. I started this PR with always and switched to on-failure, I thought I'd checked it but must have misremembered. Shall I send a PR to change it to always in buildbot-nix?

modules/nixos/buildbot.nix
Mic92: Afaik it reconnects eventually.

PR #1241, opened 2024-04-27
flake: switch back to upstream nixpkgs-update

zowoq:
Currently using my fork, need to fix and merge a few upstream PRs before we can switch back.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -142,21 +142,21 @@
       "inputs": {
         "mmdoc": [],
         "nixpkgs": "nixpkgs_2",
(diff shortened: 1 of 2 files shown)


flake.lock
```diff
@@ -228,6 +228,22 @@
         "treefmt-nix": "treefmt-nix"
       }
     },
+    "runtimeDeps": {
```
zowoq, line 231: This input will be dropped after the upstream flake is updated in https://github.com/nix-community/nixpkgs-update/pull/419.

PR #1252, opened 2024-05-02
modules/nixos: add disko-zfs-systemd-boot

zowoq:


Mic92 (approved):

Mic92:
Should I test installing this on build01?

zowoq:
> Should I test installing this on build01?

No, I'll do it myself.

I would appreciate a response to https://github.com/nix-community/infra/pull/1207#issuecomment-2076193777.

Mic92:
> > Should I test installing this on build01?
> 
> No, I'll do it myself.
> 
> I would appreciate a response to [#1207 (comment)](https://github.com/nix-community/infra/pull/1207#issuecomment-2076193777).

Without trying it myself, I can't tell.

zowoq:
@Mic92 I tried this on build01. Installed, booted, another zfs import error. Is there some sort of weird systemd-boot / zfs problem?

Mic92:
> @Mic92 I tried this on build01. Installed, booted, another zfs import error. Is there some sort of weird systemd-boot / zfs problem?

No. I use systemd-boot in many places. Only thing I can think of is that the pool is not exported correctly. In nixos-anywhere we do this: https://github.com/nix-community/nixos-anywhere/blob/f99d120b3788a286989db4e592a698f5d310d2f6/src/nixos-anywhere.sh#L516

zowoq:
I disabled uefi on build03 so now it is the same as 01 and 02.

modules/nixos/disko-zfs-swap.nix
```diff
@@ -0,0 +1,86 @@
...
+  options = {
+    nixCommunity.disko.swapSize = lib.mkOption {
+      type = lib.types.int;
+      default = null;
+      description = "swap size in GB";
+    };
+  };
+
+  config = {
+    networking.hostId = "deadbeef";
```
Mic92, line 51:
```suggestion
    networking.hostId = "8425e349"; # same as nixos-anywhere
```

Mic92: ^ than we never have to force import a pool.

Mic92: This is now also port of srvos, I believe.

modules/nixos/disko-zfs-zvol.nix
```diff
@@ -0,0 +1,97 @@
...
+            xattr = "sa";
+            "com.sun:auto-snapshot" = "false";
+          };
+          datasets = {
+            root = {
+              type = "zfs_fs";
+              mountpoint = "/";
+              options.mountpoint = "legacy";
+            };
+            writeback = {
```
zowoq, line 88:
```
cannot create 'zroot/writeback': out of space
```

Mic92: For a new installation or incremental one?

modules/nixos/disko-zfs-zvol.nix
```diff
@@ -0,0 +1,97 @@
...
+    boot.loader.grub = {
+      enable = true;
+      efiSupport = true;
+      efiInstallAsRemovable = true;
+    };
+
+    # Sometimes fails after the first try, with duplicate pool name errors
+    boot.initrd.systemd.services.zfs-import-zroot.serviceConfig.Restart = "on-failure";
+
+    zramSwap.writebackDevice = "/dev/zvol/zroot/writeback";
```
Mic92, line 61: Doesn't this suffer from the same issue that we also have with swap files on zfs filesystems? Also now that zfs has proper direct I/O maybe this is not an issue? https://github.com/openzfs/zfs/releases/tag/zfs-2.3.0-rc1

modules/nixos/disko-zfs-zvol.nix
zowoq: New. 

Disko installTest log is here: https://buildbot.nix-community.org/#/builders/3544/builds/1/steps/1/logs/stdio.

modules/nixos/disko-zfs-zvol.nix
zowoq: zvol is used as an example in the zram-generator docs so I'm assuming that it works.

https://github.com/systemd/zram-generator/blob/766bfb0d297fe083cc29ab1886d8feff24f2ee58/man/zram-generator.conf.md?plain=1#L85

modules/nixos/disko-zfs-zvol.nix
zowoq: Seems to be a problem with the size of the disk, test works if it is 1G or created as a sparse disk. I've seen some comments that sparse can't be problematic so not going to use it.

zowoq:
> Going to test this on build01, created the volume manually and added the writeback device in #1689.

Seems that doesn't do anything, it doesn't have any defaults, users need to write their own triggers and set timers for the triggers. Reverted in https://github.com/nix-community/infra/pull/1696.

PR #1255, opened 2024-05-05
add agenix to deploy darwin secrets, darwin03: convert to CI builder

zowoq:
I had been thinking of implementing nix-darwin support in sops-nix but current the only use I would have for it is here.

Rather than continuing to manage secrets manually on darwin we may as well use agenix.

Not tested yet, thought I'd get feedback on it first.


zowoq:
As there isn't any immediate benefit from merging so I'll leave it as a draft for now while the sops-nix PR is in progress, we can merge this PR if we do need to touch secrets on the darwin.

PR #1285, opened 2024-05-26
darwin/community-builder: add key for imincik

imincik:
Sorry for another PR, but I need to add one more SSH key which is not password protected and will work for Nix remote builder.

Thank you.


imincik:
@zowoq , great thanks for so fast response time ! Have a nice day.

PR #1287, opened 2024-05-27
docs: refactor 

zowoq:


zimbatm:
I'd love to see some motivating message with the refactor PR. When looking at those it's not always clear to me why the changes are needed, or what lessons were learned. Bystanders could also benefit from the learned wisdom.

dev/docs.nix
```diff
@@ -23,5 +23,19 @@
         cd $files
         mkdocs build --strict --site-dir $out
       '';
+    docs-linkcheck = pkgs.testers.lycheeLinkCheck rec {
```
zowoq, line 26: Previously I hadn't been to concerned with implementing linkchecking but as this tester is in nixpkgs it make the maintenance burden of it very low so we may as well use it.

PR #1291, opened 2024-05-29
darwin & linux community builder: add a sane zsh config

Artturin:
Biggest benefits are better prompt and the ability to properly use arrow up
to go through the history

`nix<up>` searches through commands which start with `nix`

https://grml.org/

I like using this as the the default configuration on all my machines

### DIFF
--- a/modules/darwin/community-builder/default.nix
+++ b/modules/darwin/community-builder/default.nix
@@ -1,6 +1,16 @@
+{ pkgs, ... }:
 {
   imports = [
     ./packages.nix
     ./users.nix
   ];
+
+  programs.zsh = {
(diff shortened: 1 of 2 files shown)


modules/darwin/community-builder/default.nix
```diff
@@ -1,6 +1,16 @@
+{ pkgs, ... }:
 {
   imports = [
     ./packages.nix
     ./users.nix
   ];
+
+  programs.zsh = {
```
zowoq, line 8: Moved this from darwin/common to darwin/community-builder. 

Zsh is only enabled on all the darwin machines because is the the macos default.

PR #1298, opened 2024-06-01
nix-topology

zowoq:


Mic92:
cool idea.

zowoq:
Currently it isn't very useful for us, e.g. doesn't have support for nix-darwin, need to see if upstream is open to adding that.

zowoq:
Leaving this for now. Doesn't really support the density of information that I'd like to include (e.g.  https://github.com/nix-community/infra/pull/1673) but may revisit if I have a use for it elsewhere.

PR #1304, opened 2024-06-08
modules/nixos/buildbot: switch to github app

zowoq:


Mic92:
Depends on https://github.com/Mic92/buildbot-nix/pull/156

Mic92:
Maybe wait a couple of days with it and I have verified keys are no longer expiring.

Mic92:
This should be the last bug fixed: https://github.com/nix-community/buildbot-nix/pull/188
It was stress tested with scripts afterwards.

zowoq:
I'll wait a few days and then do a clean up, previous oauth app, previous buildbot user permissions, etc.

zowoq:
I demoted the nix-community-buildbot user from owner to member after we switch to the app but now I've removed it from the org. It's token has also been revoked. The original buildbot oauth app has been deleted.

Now that buildbot.numtide and buildbot.thalheim have both switch to the github app I've removed the numtide-bot user account as a collaborator and removed the original oauth permission that was used for buildbot.thalheim.

numtide

thalheim

PR #1319, opened 2024-06-20
modules/darwin: authorizedKeys updates

zowoq:


modules/darwin/common/default.nix
```diff
@@ -58,15 +57,11 @@ in
   # disable application layer firewall, telegraf needs an incoming connection
   system.defaults.alf.globalstate = 0;
 
-  # srvos
-  services.openssh.authorizedKeysFiles = pkgs.lib.mkForce [ "/etc/ssh/authorized_keys.d/%u" ];
-
   # srvos
   environment.etc."ssh/sshd_config.d/darwin.conf".text = ''
     HostKey /etc/ssh/ssh_host_ed25519_key
```
zowoq, line 62:
```suggestion
  environment.etc."ssh/sshd_config.d/darwin.conf".text = ''
    AuthorizedKeysFile none
    HostKey /etc/ssh/ssh_host_ed25519_key
```

emilazy:
Just a heads up: https://github.com/LnL7/nix-darwin/pull/976 has now been merged, so you shouldn’t need to use the fork here and the next upgrade would start complaining about `/etc/authorized_keys.d` anyway because of the security warning :)

modules/darwin/common/default.nix
zowoq: We may need to reinstall or switch the community builder to a different machine so easier to just require that the keys are managed.

emilazy: Makes sense, I forgot about the warning about `$HOME` :)

PR #1335, opened 2024-07-02
hydra: build CUDA packages for the CUDA team

zimbatm:
I don't know how much resources that would take, but we could give it a go.

CUDA packages are slow to build, and not cached by upstream due to upstream not building unfree packages. This could help the team quite a bit.

### DIFF
--- a/terraform/hydra-projects.tf
+++ b/terraform/hydra-projects.tf
@@ -69,3 +69,55 @@ resource "hydra_project" "simple_nixos_mailserver" {
     value = "https://gitlab.com/simple-nixos-mailserver/nixos-mailserver"
   }
 }
+
+resource "hydra_project" "nixpkgs" {
+  name         = "nixpkgs"
+  display_name = "nixpkgs"
+  description  = "you know what this is"
+  homepage     = "https://github.com/NixOS/nixpkgs"
+  owner        = "admin"
+  enabled      = true
+  visible      = true
+}
+
+resource "hydra_jobset" "nixpkgs_cuda" {
+  project     = hydra_project.nixpkgs.name
+  state       = "enabled"
+  visible     = true
+  name        = "cuda"
+  type        = "legacy"
+  description = "Testing CUDA support. Come help the CUDA team! https://nixos.org/community/teams/cuda/"
+
+  nix_expression {
+    file  = "pkgs/top-level/release-cuda.nix"
+    input = "nixpkgs"
+  }
+
+  input {
+    name              = "nixpkgs"
+    type              = "git"
+    value             = "https://github.com/NixOS/nixpkgs.git nixos-unstable-small"
+    notify_committers = false
+  }
+
+  input {
+    name              = "officialRelease"
+    type              = "boolean"
+    value             = "false"
+    notify_committers = false
+  }
+
+  input {
+    name              = "supportedSystems"
+    type              = "nix"
+    value             = "[ \"x86_64-linux\" ]"
+    notify_committers = false
+  }
+
+  check_interval    = 1800
+  scheduling_shares = 3000
+  keep_evaluations  = 1
+
+  email_notifications = false
+  email_override      = ""
+}


terraform/hydra-projects.tf
```diff
@@ -69,3 +69,62 @@ resource "hydra_project" "simple_nixos_mailserver" {
...
+resource "hydra_jobset" "nixpkgs_cuda" {
+  project     = hydra_project.nixpkgs.name
+  state       = "enabled"
+  visible     = true
+  name        = "cuda"
+  type        = "legacy"
+  description = "Testing CUDA support. Come help the CUDA team! https://nixos.org/community/teams/cuda/"
+
+  nix_expression {
+    file  = "pkgs/top-level/release-cuda.nix"
```
SomeoneSerge, line 92: Uh-huh, we'd have to update that in Nixpkgs, this currently just lists a small and arbitrary selection of packages

terraform/hydra-projects.tf
SomeoneSerge, line 1: As simple as that 🙃. This would be really great, who are the parties we need to negotiate with?

zimbatm: @zowoq because he is doing almost all of the maintenance work on the nix-community infra.

terraform/hydra-projects.tf
zimbatm: Yeah, that will be up to you. I was wondering the same about the supportedPlatforms. Maybe that's something you want to control on the other end as well?

terraform/hydra-projects.tf
```diff
@@ -69,3 +69,62 @@ resource "hydra_project" "simple_nixos_mailserver" {
...
+    input = "nixpkgs"
+  }
+
+  input {
+    name              = "limitedSupportedSystems"
+    type              = "nix"
+    value             = "[ ]"
+    notify_committers = false
+  }
```
zimbatm, line 101:
```suggestion
  input {
    name              = "limitedSupportedSystems"
    type              = "nix"
    value             = "[ ]"
    notify_committers = false
  }
```
The file doesn't take a limitedSupportedSystems so can probably be dropped?

terraform/hydra-projects.tf
```diff
@@ -69,3 +69,62 @@ resource "hydra_project" "simple_nixos_mailserver" {
...
+
+  input {
+    name              = "nixpkgs"
+    type              = "git"
+    value             = "https://github.com/NixOS/nixpkgs.git nixos-unstable"
```
zimbatm, line 106: @SomeoneSerge do you want to use another branch, like a staging-cuda or something?

terraform/hydra-projects.tf
SomeoneSerge: Well supported platforms are just `*-linux` at this point, I think it's ok to keep that constant

terraform/hydra-projects.tf
SomeoneSerge: We tried pushing big changes to `cuda-updates` first and building it prior to merging into `master`, but we kept coming back to targeting `master` directly. In our [hercules](https://hercules-ci.com/github/SomeoneSerge/nixpkgs-cuda-ci) we build master + nixos-unstable + the latest release: this way by the time CI starts a round of nixos-unstable, some of will have been cached by the master job. This alleviates some of the pain of `nixos-unstable` advancing without testing CUDA

zowoq:
> I don't know how much resources that would take, but we could give it a go.

No objection. Maybe if this ends up too much for our current hardware we could consider upgrading.

terraform/hydra-projects.tf
```diff
@@ -69,3 +69,62 @@ resource "hydra_project" "simple_nixos_mailserver" {
...
+  input {
+    name              = "supportedSystems"
+    type              = "nix"
+    value             = "[ \"x86_64-linux\" \"aarch64-linux\" \"aarch64-darwin\" ]"
+    notify_committers = false
+  }
+
+  check_interval    = 1800
+  scheduling_shares = 3000
+  keep_evaluations  = 3
```
zowoq, line 126: Everything is pushed to cachix so only need to keep the current evaluation.
```suggestion
  keep_evaluations  = 1
```

Mic92:
> > I don't know how much resources that would take, but we could give it a go.
> 
> No objection. Maybe if this ends up too much for our current hardware we could consider upgrading.

If you tell companies that all they need to do to get CUDA packages for NixOS is to pay some monthly donations, than I am sure we get the funding pretty quick.

terraform/hydra-projects.tf
```diff
@@ -114,13 +107,6 @@ resource "hydra_jobset" "nixpkgs_cuda" {
     notify_committers = false
   }
 
-  input {
-    name              = "supportedSystems"
-    type              = "nix"
-    value             = "[ \"x86_64-linux\" \"aarch64-linux\" \"aarch64-darwin\" ]"
```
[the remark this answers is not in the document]
zimbatm: For now it's "x86_64-linux" only since upstream has set that value as a default.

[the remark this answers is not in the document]
zowoq: Yes, I mean we should restrict it here to linux only as well.

zimbatm:
First run: https://hydra.nix-community.org/eval/109915

terraform/hydra-projects.tf
```diff
@@ -81,10 +81,12 @@ resource "hydra_project" "nixpkgs" {
 }
 
 resource "hydra_jobset" "nixpkgs_cuda" {
+  for_each = toset(["nixos-unstable", "master"])
```
zowoq, line 84: 👎  on master, nixos-unstable small?

zimbatm: do we even need nixos-unstable if nixos-unstable-small is targeted?

zowoq: I don't think so, just small should work?

terraform/hydra-projects.tf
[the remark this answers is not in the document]
SomeoneSerge: We support both x86_64 and aarch64 linux, I'll update the release file

terraform/hydra-projects.tf
SomeoneSerge, line 1: @zowoq so I'm considering just building the whole `import <nixpkgs> { config.cudaSupport = true; }`; this gives me something like

```
❯ nix-eval-jobs --expr 'import ./pkgs/top-level/release-cuda.nix { }' --force-recurse | wc -l
...
# eval errors, eval errors
...
138452
```

Does that sound unreasonable? I could in principle come up with a smaller, curated set of jobs.

Hexa also raises the concern that this would be effectively mirroring the NixOS Hydra:

> hexa (UTC+1)
> so all of them
> if there was a cache behind nix-community hydra, than you'd be mirroring cache.nixos.org effectively
> SomeoneSerge (UTC+3)
> Yeah... Ideally we'd have a solution that evaluates the full DAGs for vanilla and cuda nixpkgs, starts building cuda from the leaves (ehhh, the roots), and always suspends the build if it hash matches the vanilla hash

zowoq: > `import <nixpkgs> { config.cudaSupport = true; }`

TBH I doubt we have the capacity to handle that without it being detrimental to the other projects using this CI. Currently we only have two builders for linux that are shared by buildbot, hercules, hydra:

```
### `build03`

- Provider: Hetzner
- CPU: AMD Ryzen 9 3900 12-Core Processor
- RAM: 128GB DDR4 ECC
- Drives: 2 x 1.92 TB NVME in RAID 1

### `build04`

- Provider: Hetzner
- Instance type: [RX170](https://www.hetzner.com/dedicated-rootserver/rx170)
- CPU: Ampere Altra Q80-30 80-Core Processor
- RAM: 128GB DDR4 ECC
- Drives: 2 x 960 GB NVME in RAID 0
```

If you do want to build everything maybe we could have dedicated machines just for this package set? Not sure if raising the money for that is feasible?

> if there was a cache behind nix-community hydra, than you'd be mirroring cache.nixos.org effectively

Not sure if I've misunderstood or not, we push everything to cachix but it skips existing nixos cache paths.

SomeoneSerge: > Not sure if I've misunderstood or not, we push everything to cachix but it skips existing nixos cache paths.

The concern is that if there is a phase shift between NixOS and the Community Hydras, and the _latter_ starts building a certain derivation from a certain commit _before_ the former, we'll have wasted some storage and compute

SomeoneSerge: > TBH I doubt we have the capacity to handle that without it being detrimental to the other projects using this CI. Currently we only have two builders for linux that are shared by buildbot, hercules, hydra:

Roger that. I'll push a smaller jobset tomorrow, based on what we've been building in https://github.com/SomeoneSerge/nixpkgs-cuda-ci.

How id the community builder funded? @ConnorBaker was asking on matrix if there's an opencollective yet

zowoq: > The concern is that if there is a phase shift between NixOS and the Community Hydras, and the _latter_ starts building a certain derivation from a certain commit _before_ the former, we'll have wasted some storage and compute

Yeah, we can't really avoid that with this approach. Could try adding the free deps as blockers for nixos-unstable or could try doing something similar in a repo here, flake update PRs with `max-jobs = 0` so merging is blocked they aren't cached?

> How is the community builder funded? ConnorBaker was asking on matrix if there's an opencollective

We have an opencollective: https://opencollective.com/nix-community

ConnorBaker:
> > > I don't know how much resources that would take, but we could give it a go.
> > 
> > 
> > No objection. Maybe if this ends up too much for our current hardware we could consider upgrading.
> 
> If you tell companies that all they need to do to get CUDA packages for NixOS is to pay some monthly donations, than I am sure we get the funding pretty quick. Also we still haven't asked Hetzner for the discount they are offering to the NixOS foundation. This way we would probably still save money with bigger hardware.

They offer discounts? My Hetzner bill for part of the CUDA CI is like $400; I’d love to consolidate some of that stuff under the community, especially if you can get a discount and we can all benefit from it!

terraform/hydra-projects.tf
zimbatm: Ok, fixed in e310b5d104de3254f017ccac4718f2ac6e04fc27

terraform/hydra-projects.tf
zimbatm: A related idea would be to pull from nix-community/nixpkgs and give the CUDA team access to it. That would limit the risks compared to giving everybody push access to NixOS/nixpkgs

zimbatm: For now we're building nixos-unstable-small

zimbatm:
Merging the current state. We can still do follow-up PRs afterwards!

zimbatm:
> They offer discounts? My Hetzner bill for part of the CUDA CI is like $400; I’d love to consolidate some of that stuff under the community, especially if you can get a discount and we can all benefit from it!

Yes, but they ran out of the discount budget for this year. We'll have to contact them again next year.

zimbatm:
If you want to donate hardware to the cause, we are discussing what the requirements would be in https://github.com/nix-community/infra/issues/1343

zowoq:
> My Hetzner bill for part of the CUDA CI is like $400

Could you go into some detail please? What hardware, what is built and what is the utilization like?

zowoq:
I've reverted this as it had been interfering with our other CI builds.

I'll see if I can find a way of running these builds without causing problems for our other users.

zimbatm:
The easiest would be to build a second cluster. I'll see if we can raise another 2-300/month for this.

PR #1337, opened 2024-07-03
sops-nix -> agenix

zowoq:
Any objections to moving to agenix for deployed secrets? 

We'd still be using sops for two files: `/secrets.yaml` and `/terraform/secrets.yaml`.


zimbatm:
Ideally, we would use one tool everywhere. I don't care either way.

Is that your motivation to make the change?

zowoq:
After using both for a while I just prefer it as it is simpler.

zimbatm:
No objection. If we need KMS support down the line, we can always switch back.

zowoq:
> After using both for a while I just prefer it as it is simpler.

Ended up that I prefer neither. I'm looking at other options but running a vault service is a bit excessive for our current needs and some of the other tools are very new so for now it is easier to go back to sops-nix.

PR #1340, opened 2024-07-04
terraform/hydra-projects: add cuda, rocm

zowoq:


Mic92:
Would be good to document what we do need to make this work, so we can reach out to the community to get more build resources.

zowoq:
Don't really need more resources, these builds are heavy while they are running but don't add that much overall load. The issue is that for the few hours while they are building they are competing for resources with the buildbot builds.

At moment I'm thinking of moving the localhost hydra builder into a nixos container and capping it at 50% cpu/memory, I don't think there are many options to limit only some nix builds and not others.

SomeoneSerge:
When building packages with CUDA support for multiple CUDA architectures you usually want to set `nix.settings.max-jobs` to a lower value (e.g. `nix-build ... -j1`), otherwise even simple packages like `faiss` spawn a dozen `ptxas` processes at linkage time, consume all CPUs and cause an OOM. I've been also using `services.hercules-ci-agent.settings.concurrentTasks = 2` and `zramSwap.enable`.

To address this properly we need some kind of negative affinity in the nix build scheduler

Mic92:
> Don't really need more resources, these builds are heavy while they are running but don't add that much overall load. The issue is that for the few hours while they are building they are competing for resources with the buildbot builds.
> 
> At moment I'm thinking of moving the localhost hydra builder into a nixos container and capping it at 50% cpu/memory, I don't think there are many options to limit only some nix builds and not others.

I just played a bit around. I think there is even an easier solution: Having a different nix-daemon:

```
/run/systemd/system/nix-daemon2.service
[Unit]
Description=Nix Daemon
Documentation=man:nix-daemon https://nixos.org/manual
RequiresMountsFor=/nix/store
RequiresMountsFor=/nix/var
RequiresMountsFor=/nix/var/nix/db
ConditionPathIsReadWrite=/nix/var/nix/daemon-socket

[Service]
ExecStart=@/nix/store/riq2rqg9c5i5687m5h3m1qhlzci9chb0-nix-2.23.2/bin/nix-daemon nix-daemon --daemon
Environment=NIX_DAEMON_SOCKET_PATH=/run/hydra-nix.sock
KillMode=process
LimitNOFILE=1048576
TasksMax=1048576

[Install]
WantedBy=multi-user.target
```

```
NIX_DAEMON_SOCKET_PATH=/run/hydra-nix.sock strace -e connect -f nix-build -A hello
strace: Process 571584 attached
[pid 571583] connect(3<UNIX-STREAM:[1819608]>, {sa_family=AF_UNIX, sun_path="/run/hydra-nix.sock"}, 110) = 0
```

Than we can use systemd to apply cpu shares or we can set different cpu priorities.

This would also allows us to have different build core settings.

Mic92:
Having different nix-daemons works because all locks that nix needs, also work across processes because when you run nix as root, it actually directly modifies the database rather than talking to the nix-daemon.

zowoq:
How about a different daemon on a remote builder? Eventually we'll have the same problems on the aarch64 machine, I was thinking if the container idea does work we could have, e.g. 2 containers on build03,build04 and the containers are each capped at 25%?

Mic92:
> How about a different daemon on a remote builder? Eventually we'll have the same problems on the aarch64 machine, I was thinking if the container idea does work we could have, e.g. 2 containers on build03,build04 and the containers are each capped at 25%?

I think a different nix-daemon would also work with remote builder. It would require a nix remote builder user with the `NIX_DAEMON_SOCKET_PATH` environment variable set. This can for example set by having a wrapped login shell that always export the environment variable: [users.users.<name>.shell](https://search.nixos.org/options?channel=24.05&show=users.users.%3Cname%3E.shell&from=0&size=50&sort=relevance&type=packages&query=users.users).
This should also work than on macOS, where don't have containers and where disk space is more of an issue. I also would instead of capping the CPU to 25%, I would apply a CPU share:
https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/6/html/resource_management_guide/sec-cpu_and_memory-use_case#sec-cpu_and_memory-use_case
This allows a build to use all CPU when no one else needs it.

Mic92:
I think also just the fact there are two different cgroups already should split CPU shares 50%/50%

SomeoneSerge:
You can also do something explicit like `systemd.slices.jhub.sliceConfig.CPUQuota = "50%";` I believe

zowoq:
Different nix-daemon is simpler than the container so I'll try that first on build03 and see how it goes.

Mic92:
> CPUQuota

This setting doesn't allow a process to use 100% cpu if no one else uses the system. This seems a bit of a waste.

zowoq:
@ConnorBaker @SomeoneSerge Could you suggest a few of the most resource hungry packages in the cuda package set that I can use for testing please?

SomeoneSerge:
> Could you suggest a few of the most resource hungry packages in the cuda package set that I can use for testing please?

`python3Packages.tensorflow`, `python3Packages.pytorch`, `python3Packages.jaxlib`, `python3Packages.opencv4`; all with `{ config.cudaSupport = true }`; for a shorter bootstrap chain, `faiss`

zowoq:
Sorry, I forgot to post a progress report.

Throttling via max-jobs isn't great but seems to be enough at it'll do as an interim solution while I'm working on a proper fix.

I'll merge this as is and continue work in another PR.

zowoq:
These jobsets stopped evaluating a few days ago, not sure if it needs to be fixed here or in nixpkgs.

zimbatm:
It looks like nixos-unstable is stuck for the past 3 days:
image

zowoq:
The jobsets are on nixos-unstable-small not nixos-unstable or nixpkgs-unstable.

SomeoneSerge:
Seems like the actual hydra is passing more flags. At the least I need to add `"nixpkgs"` to the list in `{ inherit supportedSystems nixpkgsArgs; } // builtins.removeAttrs args [ "variant" ]`

```

       error: function 'anonymous lambda' called with unexpected argument 'nixpkgs'
       at /nix/store/rqh0dgkndnbawlnjd3j9n9906gr2ima1-source/pkgs/top-level/release-lib.nix:1:1:
            1| { supportedSystems
             | ^
            2| , system ? builtins.currentSystem
error: worker error: error:
              … in the left operand of the update (//) operator
                at /nix/store/rqh0dgkndnbawlnjd3j9n9906gr2ima1-source/pkgs/top-level/release-cuda.nix:163:5:
                 162|     }
                 163|     // (lib.genAttrs packageSets evalPackageSet);
                    |     ^
                 164| in

              … from call site
                at /nix/store/rqh0dgkndnbawlnjd3j9n9906gr2ima1-source/pkgs/top-level/release-cuda.nix:56:17:
                  55| let
                  56|   release-lib = import ./release-lib.nix (
                    |                 ^
                  57|     { inherit supportedSystems nixpkgsArgs; } // builtins.removeAttrs args [ "variant" ]
```

(from https://hydra.nix-community.org/jobset/nixpkgs/cuda#tabs-errors)

SomeoneSerge:
H'mm in future, how do we recover from mistakes like these without waiting for master to reach unstable-small?

zowoq:
> H'mm in future, how do we recover from mistakes like these without waiting for master to reach unstable-small?

We could manually trigger an eval of a PR branch before it is merged.

zowoq:
https://hydra.nix-community.org/eval/43199, https://hydra.nix-community.org/eval/43201

That works, thanks. Checked cuda and rocm.

PR #1349, opened 2024-07-06
hydra: switch to fork

zowoq:
Manually deployed on build03 for testing.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -24,6 +24,8 @@
     flake-compat.url = "github:nix-community/flake-compat";
     flake-parts.inputs.nixpkgs-lib.follows = "nixpkgs";
     flake-parts.url = "github:hercules-ci/flake-parts";
+    hydra.flake = false;
+    hydra.url = "github:qowoz/hydra/community";
     lite-config.url = "github:yelite/lite-config";
     nix-darwin.inputs.nixpkgs.follows = "nixpkgs";
     nix-darwin.url = "github:LnL7/nix-darwin";
@@ -61,6 +63,15 @@
         {
           nixpkgs = {
             config.allowUnfreePredicate = pkg: builtins.elem (lib.getName pkg) [ "terraform" ];
+            overlays = [
+              (final: prev: {
+                hydra = (prev.hydra.override { nix = final.nixVersions.nix_2_22; }).overrideAttrs (o: {
+                  version = inputs.hydra.shortRev;
+                  src = inputs.hydra;
+                  buildInputs = o.buildInputs ++ [ final.perlPackages.DBIxClassHelpers ];
+                });
+              })
+            ];
           };
 
           hostModuleDir = ./hosts;
--- a/flake.lock
+++ b/flake.lock
@@ -147,6 +147,23 @@
         "type": "github"
       }
     },
+    "hydra": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1726454607,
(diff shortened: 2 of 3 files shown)


flake.nix
```diff
@@ -24,6 +24,8 @@
     flake-compat.url = "github:nix-community/flake-compat";
     flake-parts.inputs.nixpkgs-lib.follows = "nixpkgs";
     flake-parts.url = "github:hercules-ci/flake-parts";
+    hydra.flake = false;
+    hydra.url = "github:qowoz/hydra/community";
```
Mic92, line 28: Should we just ask for commit access and get those commits in upstream? 
We can probably even test them on the actual hydra?

zowoq: Almost all of them have been deploy deployed on hydra.nixos.org for a while and are in various upstream branches / PRs.

Mic92: cc @zimbatm @Ericson2313 @dasJ

Mic92: @zowoq even more a reason to just merge them into the hydra branch. I ask for access.

zowoq: I'm also looking at picking the lix patches that replace hydra-eval-jobs with nix-eval-jobs. Would need to pick a couple their patches for nix-eval-jobs as well.

Mic92: I am also interested in having those patches in my nix-eval-jobs to make it work with hydra.

Mindavi: I'd also like some more energy put into that hydra than various forks sprawling. But I understand it's hard since it sometimes seems nobody's really watching there.

Ericson2314: I still want to do CA things for hydra, and newer build protocol, but I have to finish the Mesonification first.

Mic92: @Ericson2314 mesonification of nix itself?

zowoq: > I'm also looking at picking the lix patches that replace hydra-eval-jobs with nix-eval-jobs.

Seems this will require a bit more work than just picking the patches so I'll leave it for now.

Mic92: @zowoq if would look at this, would you be ok, if we run the result on nix-community infra?

zowoq: I'm happy to pick patches from a upstream PR if that is what you mean?

Mic92: I mean, I would need to test patches as well in a non-trivial setup and do fixes as needed. Local testing of demo instances only take you so far.

zowoq: For that I'd suggest deploying another hydra instance on build01.

Mic92: I was hoping we could test hydra changes together, but in that case I could also look for a different project where I can deploy a hydra.

zowoq:
I've deployed the patches for this: https://github.com/NixOS/hydra/security/advisories/GHSA-xv29-v93r-2f5v.

flake.nix
zowoq: @Ericson2314 @Mic92 I'm seeing lots of new changes, any plans on merging the existing commits from hydra.nixos.org branch?

Ericson2314: @zowoq I've reviewed and merged a couple. 

Either way, they aren't getting lost. That branch gets rebased every time `hydra.nixos.org` does a new deploy.

zowoq: Thanks for the response but I am already aware of all of that and that also didn't address my question?

zowoq: Some of those commits have been in use on hydra.nixos.org for almost a year and in use here for six months.

PR #1363, opened 2024-07-11
add initrd-ssh for debugging boot issues

Mic92:


Mic92:
Merging because comin keeps reverting this change otherwise.

modules/nixos/disko-zfs.nix
```diff
@@ -13,6 +13,9 @@
     efiInstallAsRemovable = true;
   };
 
+  # Sometimes fails after the first try, with duplicate pool name errors
+  boot.initrd.systemd.services.zfs-import-zroot.serviceConfig.Restart = "on-failure";
```
zowoq, line 17: @Mic92 Should we have this enabled by default?

modules/nixos/emergency-access.nix
```diff
@@ -0,0 +1,4 @@
+{
+  # allow to access emergency shell with a password
+  boot.initrd.systemd.emergencyAccess = "$6$he2fblfl/H7I.kvz$WbSCMXu8ztmqfj5jG4czqvu/rkMHxufxqHgy1urzXFSN.jZB4QiW5lOjR08vk8pZTyim3TT1wFkMaNE9zZ3sc1";
```
zowoq, line 3: @Mic92 Should we have this enabled by default?

Mic92: Potentially. Back then, I was using `disko-install` to enable it.

Mic92: We can also enable ssh in the initrd as well so that we don't need to order a kvm switch every time.

Issue #1362, opened 2024-07-11
build02 (nixpkgs-update bot) is down


zowoq:
Stuck at boot, zfs import error, tried reinstalling, same error.

I think it is a hardware issue but not sure what the problem is yet.

Not sure how much time I'll have to work on this today.

Mic92:
Any S.M.A.R.T. values?

zowoq:
I left the hwcheck tool running for a while and it doesn't seem to have done anything.

It is in rescue mode and you key is loaded if you want to have a look at it.

Mic92:
Thanks. I am currently in parallel fixing another server. I have a look

Mic92:
Zfs support in the rescue system is broken so I did this:

curl -L https://github.com/nix-community/nixos-images/releases/download/nixos-unstable/nixos-kexec-installer-noninteractive-x86_64-linux.tar.gz | tar -xzf- -C /root
/root/kexec/run

Mic92:
Both nvme's look ok according to smartmontools:

```
Available Spare:                    100%
Available Spare Threshold:          10%
```

Mic92:
Zfs scrub is happy so far.

Mic92:
I also run out of time just now. If the problem still persist in the evening, I am going to have a look with the kvm console.

Mic92:
Ok. I will order one now.

Mic92:
All it needed was a `zpool import -f zroot`...
I think I will make the default in disko. The amount of installations that gets messed up by this is higher than the few iscsi users that might import a block devices not meant for them.

Mic92:
Ok. Weird: https://github.com/NixOS/nixpkgs/blob/249fbde2a178a2ea2638b65b9ecebd531b338cf9/nixos/modules/tasks/filesystems/zfs.nix#L650C28-L650C43
This should be already true.

Mic92:
Ok. looks like the same issue happens again after reboot. Maybe the timeout for the device is too short.

Mic92:
Now it imports the pool instant again.

zowoq:
Thanks for fixing this. I'll restore the nixpkgs-update logs from the backup.

Mic92:
Ok. I didn't had to reformat anything, but you probably did before.

PR #1364, opened 2024-07-11
different workaround for zpool import

Mic92:
fixes https://github.com/nix-community/infra/issues/1362

### DIFF
--- a/hosts/build02/configuration.nix
+++ b/hosts/build02/configuration.nix
@@ -1,4 +1,4 @@
-{ inputs, ... }:
+{ config, inputs, ... }:
 
 {
   imports = [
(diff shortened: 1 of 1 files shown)


hosts/build02/configuration.nix
```diff
@@ -12,6 +12,9 @@
     inputs.self.nixosModules.emergency-access
   ];
 
+  # the default zpool import services somehow times out while this import works fine?
+  boot.initrd.systemd.services.zfs-import-zroot.serviceConfig.ExecStartPre = "${config.boot.zfs.package}/bin/zpool import -N -f zroot";
```
Mic92, line 16: I have the feeling there are still uids somewhere that get picked up by zfs from old pools.... And than it timesout waiting for these ghost devices that no longer exist.
We have some logic in zfs that makes the system wait for all devices to be available, which this preStart by passes. However since we always need both devices to boot anyway, this shouldn't be an issue.

zowoq: Seems we need to use this everywhere now, was testing install on build01 and it wouldn't work without it. https://github.com/nix-community/infra/commit/347fbab61004ce5d9638bf3aa634d1f0ac052f40

PR #1368, opened 2024-07-13
modules/nixos/buildbot: set jobReportLimit

zowoq:


modules/nixos/buildbot.nix
```diff
@@ -41,8 +41,10 @@
     domain = "buildbot.nix-community.org";
     evalMaxMemorySize = "4096";
     evalWorkerCount = 32;
+    jobReportLimit = 0;
```
Mic92, line 44: 0 might be not ideal as we currently don't show progress beyond when all packages are finished. We plan to fix this soon and merged the current state to unblock bigger projects for now.

zowoq: I was thinking that there isn't going to be a number that is going to work for everyone so it would be simpler to just disable it for consistency across all projects.

For now I'll set this to 100 to most projects shouldn't notice this change and change it to 0 later.

modules/nixos/buildbot.nix
```diff
@@ -43,11 +43,11 @@
     evalWorkerCount = 32;
     jobReportLimit = 100;
     workersFile = config.sops.secrets.buildbot-nix-workers.path;
-    cachix = {
-      enable = true;
-      name = "nix-community";
-      auth.authToken.file = config.sops.secrets.cachix-auth-token.path;
-    };
+    #cachix = {
```
zowoq, line 46: We have watch store on this machine so we can live without the buildbot cachix integration for little while.

PR #1383, opened 2024-07-25
modules/nixos/common: move ssh keys to /var/lib, add initrd ssh

zowoq:


modules/nixos/common/security.nix
```diff
@@ -1,7 +1,16 @@
...
 {
   # Make sure that the firewall is enabled, even if it's the default.
   networking.firewall.enable = true;
 
+  # allow to access emergency shell with a password
+  boot.initrd.systemd.emergencyAccess = "$6$he2fblfl/H7I.kvz$WbSCMXu8ztmqfj5jG4czqvu/rkMHxufxqHgy1urzXFSN.jZB4QiW5lOjR08vk8pZTyim3TT1wFkMaNE9zZ3sc1";
+
+  boot.initrd.network.ssh = {
+    enable = true;
+    authorizedKeyFiles = pkgs.lib.filesystem.listFilesRecursive "${toString inputs.self}/users/keys";
```
Mic92, line 11: Since we use systemd in the initrd, I believe we need to also configure networkd here:

```
    boot.initrd.systemd.network.networks."10-uplink" = config.systemd.network.networks."10-uplink";
```

zowoq: I forgot that we'll need to set host keys for initrd as well.

zowoq: I'll merge emergency access for now and I'll do ssh later. https://github.com/nix-community/infra/pull/1401

zowoq: >  I believe we need to also configure networkd here

Looks like this is configured by default?

modules/nixos/common/security.nix
```diff
@@ -1,10 +1,32 @@
...
+    enable = true;
+    ssh = {
+      enable = true;
+      authorizedKeyFiles = lib.filesystem.listFilesRecursive "${inputs.self}/users/keys";
+      hostKeys = [ initrd_host_key ];
+    };
+  };
+
+  system.activationScripts.initrd-ssh-host-key = ''
+    [[ -f ${initrd_host_key} ]] || ${config.programs.ssh.package}/bin/ssh-keygen -q -t ed25519 -N "" -f ${initrd_host_key}
```
Mic92, line 27: That will fail on the first deployment, but than work on the second one. So probably fine?

zowoq: > That will fail on the first deployment

Could you explain please?

Mic92: activation phase is run after the boot loader is installed. So on the first deployment this will fail.

zowoq: Added the initrd key to secrets so it'll be deployed by nixos-anywhere with the host key.

Will need to deploy the key manually to the current machines.

zowoq: @Mic92 Does this look okay now?

zowoq:
Updated this to use /var/lib for the keys so we don't need move them again to use `etc.overlay`.

tasks.py
```diff
@@ -143,20 +168,30 @@ def opener(path: str, flags: int) -> Union[str, int]:
...
+                check=True,
+                stdout=fh,
+            )
+
+    decrypt(
+        "var/lib/ssh_secrets/ssh_host_ed25519_key",
+        f'["ssh_host_ed25519_key"]["{flake_attr}"]',
+    )
+    decrypt(
+        "var/lib/ssh_secrets/initrd_host_ed25519_key", '["initrd_host_ed25519_key"]'
```
Mic92, line 193: This looks better.

zowoq:
I've removed initrd ssh in https://github.com/nix-community/infra/commit/0f6cdfd9876d4ff7041c53b950a44d7f33edd314, seems to have sometimes been causing occasional initrd-nixos-copy-secrets.service failures.

PR #1389, opened 2024-07-27
modules/darwin/community-builder: add linj

jian-lin:
I am a Nixpkgs committer.  Recently, I focus on improving Emacs infra in Nixpkgs.  I would like to use this darwin machine to do testing.


jian-lin:
Thanks!

PR #1392, opened 2024-07-28
build03: update buildbot workers

zowoq:


zowoq:
@Mic92 I had bumped the number of workers to 48 but I noticed the logs were spammed with conenction failures from workers numbered >48, seems workers = threads.

Mic92:
By default it's set to the number of cores. It can configured otherwise.

zowoq:
I don't understand, it was configured?

zowoq:
I understand now what you meant, needs to be set on both sides, I assumed the worker count was passed to the worker from the config on the master.

PR #1408, opened 2024-08-11
modules/builder: gc by percent used

zowoq:


modules/shared/nix-daemon.nix
```diff
@@ -12,14 +12,14 @@ in
     settings.substituters = [ "https://nix-community.cachix.org" ];
 
     # auto-free the /nix/store
-    settings.min-free = asGB 10;
+    settings.min-free = asGB 1;
```
Mic92, line 15: 1GB is ok? Wouldn't that easily trigger build failures if a build unpacks more than a 1G?

zowoq: Afraid I don't understand? This is more than the srvos default?

```
nix.settings.min-free = lib.mkDefault (512 * 1024 * 1024);
```

zowoq: We're likely just cargo culting min/max free anyway, only way it'll be used if the gc service fails?

PR #1412, opened 2024-08-16
darwin/community-builder: add user sebtm

SebTM:
I'm nixpkgs/home-manager maintainer, recently trying to use/help out with https://github.com/tstat/raspberry-pi-nix and attempts to build/improve NixOS on the BPI-R3 why I would like to request access to the shared aarch64-builder. Also I maintain a few packages, some of them build and work on darwin and I have no building possibility.


zowoq:
> recently trying to use/help out with https://github.com/tstat/raspberry-pi-nix and attempts to build/improve NixOS on the BPI-R3 why I would like to request access to the shared aarch64-builder.

If you are looking for access to an aarch64-linux that is in https://github.com/NixOS/aarch64-build-box.

SebTM:
Thanks, didn't know about that - will request access there 🙏🏻

PR #1426, opened 2024-08-26
modules/nixos: remove comin,add update-host

zowoq:


modules/nixos/common/update.nix
```diff
@@ -0,0 +1,52 @@
...
+
+      nix-store --realise $system
+      nix-env --profile /nix/var/nix/profiles/system --set $system
+      /nix/var/nix/profiles/system/bin/switch-to-configuration boot
+
+      booted="$(readlink /run/booted-system/{initrd,kernel,kernel-modules} && cat /run/booted-system/kernel-params)"
+      p="$(readlink -f /nix/var/nix/profiles/system)"
+      built="$(readlink $p/{initrd,kernel,kernel-modules} && cat $p/kernel-params)"
+      if [[ "$booted" != "$built" ]]; then
+        # don't use kexec if system is virtualized
```
Mic92, line 36: I guess this is the reason?

```suggestion
        # don't use kexec if system is virtualized, reboots are fast enough
```

modules/nixos/buildbot.nix
```diff
@@ -42,6 +42,7 @@
     evalMaxMemorySize = 4096;
     evalWorkerCount = 32;
     jobReportLimit = 100;
+    outputsPath = "/var/www/buildbot/nix-outputs/";
```
Mic92, line 45: I discovered that different flakes can override each other. So I wouldn't merge this until this issue is fixed (it would be a security vulnerability.

modules/nixos/common/update.bash
```diff
@@ -0,0 +1,33 @@
...
+    echo "--- diff to current-system"
+    nvd diff /run/current-system "$p"
+    echo "---"
+  fi
+  /nix/var/nix/profiles/system/bin/switch-to-configuration boot
+  # don't use kexec if system is virtualized, reboots are fast enough
+  if ! systemd-detect-virt -q; then
+    kexec --load "$p"/kernel --initrd="$p"/initrd --append="$(cat "$p"/kernel-params) init=$p/init"
+  fi
+  if [[ ! -e /run/systemd/shutdown/scheduled ]]; then
```
Mic92, line 28: Shouldn't it than also create this file?

zowoq: The file is created by systemd when a shutdown/reboot is scheduled.

PR #1433, opened 2024-09-02
modules/nixos/common: optimise zram

zowoq:


modules/nixos/common/default.nix
```diff
@@ -40,6 +40,15 @@
 
   zramSwap.enable = true;
 
+  # https://github.com/NixOS/nixpkgs/pull/268121
+  # https://wiki.archlinux.org/title/Zram#Optimizing_swap_on_zram
+  boot.kernel.sysctl = {
+    "vm.page-cluster" = 0;
```
Mic92, line 46: Are we using zstd in our case? The linked pull request does set this value conditionally.

zowoq: Yes, zstd is the default.

Mic92 (approved):
Don't have personal experience with these defaults but the linked sources look legit.

sedlund:
https://github.com/LeSuisse/nixpkgs/blob/afd3a57b83977a2e0fe8c4685ca2016a12103d03/pkgs/os-specific/linux/kernel/common-config.nix#L879

PR #1435, opened 2024-09-03
terraform/hydra-projects: add cuda-stable

mweinelt:
CUDA is one of the more costly ecosystems for users to build, so providing a cache for NixOS stable would be very much appreciated.

### DIFF
--- a/terraform/hydra-projects.tf
+++ b/terraform/hydra-projects.tf
@@ -122,6 +122,48 @@ resource "hydra_jobset" "nixpkgs_cuda" {
   email_override      = ""
 }
 
+resource "hydra_jobset" "nixpkgs_cuda_stable" {
+  project     = hydra_project.nixpkgs.name
+  state       = "enabled"
+  visible     = true
+  name        = "cuda-stable"
+  type        = "legacy"
+  description = "Testing CUDA support. Come help the CUDA team! https://nixos.org/community/teams/cuda/"
+
+  nix_expression {
+    file  = "pkgs/top-level/release-cuda.nix"
+    input = "nixpkgs"
+  }
+
+  input {
+    name              = "nixpkgs"
+    type              = "git"
+    value             = "https://github.com/NixOS/nixpkgs.git nixos-24.05-small"
+    notify_committers = false
+  }
+
+  input {
+    name              = "officialRelease"
+    type              = "boolean"
+    value             = "false"
+    notify_committers = false
+  }
+
+  input {
+    name              = "supportedSystems"
+    type              = "nix"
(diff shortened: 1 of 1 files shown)


zowoq:
https://hydra.nix-community.org/jobset/nixpkgs/cuda-stable

mweinelt:
Thanks for the fixup and the quick deploy!

zowoq:
Is this useful enough as is or do the changes to `release-cuda` need to be backported?

mweinelt:
I think backporting these changes would be useful, but I'll defer to @SomeoneSerge for that.

SomeoneSerge:
Yes, the release file in the release branch doesn't contain anything useful. We should backport the whole thing and it's safe to do so

terraform/hydra-projects.tf
```diff
@@ -122,6 +122,48 @@ resource "hydra_jobset" "nixpkgs_cuda" {
...
+    name              = "officialRelease"
+    type              = "boolean"
+    value             = "false"
+    notify_committers = false
+  }
+
+  input {
+    name              = "supportedSystems"
+    type              = "nix"
+    value             = "[ \"x86_64-linux\" ]"
```
mweinelt, line 155:
```suggestion
    value             = "[ \"x86_64-linux\" \"aarch64-linux\" ]"
```

SomeoneSerge: Note that this currently targets sbsa instead of jetsons

mweinelt: Just making sure we reach parity with unstable.

mweinelt:
The release-24.05 has received the relevant updates to `pkgs/top-level/release-cuda.nix`. Anything blocking the merge?

terraform/hydra-projects.tf
zowoq: > Note that this currently targets sbsa instead of jetsons

What needs to be changed for that, something here or in nixpkgs?

mweinelt: Hydra does not control what gets built beyond the entrypoint that it uses, so the change has to be in nixpkgs.

zowoq: Okay, lets wait for that then.

SomeoneSerge: Yes, the change has to be on the nixpkgs side, we have to supply a different list of `cudaCapabilities` for the `aarch64-linux` package set. IIRC it seemed like it's not completely straightforward to do with `release.nix`. Just got to sit down for a moment and do it

mweinelt:
Dropped aarch64-linux, because I don't expect the jetson changes to land anytime soon.

PR #1436, opened 2024-09-04
modules/shared/nix-daemon: disable channels and set flake registry, nix path

zowoq:


zimbatm:
Isn't that replicating what has been added in https://github.com/nix-community/srvos/pull/496 ?

zowoq:
> Isn't that replicating what has been added in [nix-community/srvos#496](https://github.com/nix-community/srvos/pull/496) ?

Partially, setting them here so it is one explicit implementation that overwrites the default and works on both darwin/nixos.

zimbatm (approved):

PR #1441, opened 2024-09-09
build01: enable etc.overlay and userborn

zowoq:
@Mic92 Shall we try these for a while on one host and then maybe enable them on all the hosts?

### DIFF
--- a/hosts/build01/default.nix
+++ b/hosts/build01/default.nix
@@ -7,6 +7,12 @@
     inputs.self.nixosModules.community-builder
   ];
 
+  system.etc.overlay = {
+    enable = true;
+    mutable = false;
+  };
+  services.userborn.enable = true;
+
   # Emulate riscv64 until we have proper builders
   boot.binfmt.emulatedSystems = [ "riscv64-linux" ];
 


Mic92:
Two things I noticed when using it on my laptop:
It doesn't migrate old mutable state into the writable overlay (might be not an issue here).
I had some *busy mount* errors on upgrade. I think what sometimes happens that the mount point is still accessed by processes. I think some lazy unmount option is needed. This might be a problem. But I guess you could test it.

zowoq:
> It doesn't migrate old mutable state into the writable overlay (might be not an issue here). 

Yes, shouldn't be an issue here.

> I had some _busy mount_ errors on upgrade. I think what sometimes happens that the mount point is still accessed by processes. I think some lazy unmount option is needed.

Does https://github.com/NixOS/nixpkgs/commit/7845b89e24ff9e91e6e13f66ee5a1a73958d22fb address that?

Mic92:
> Does [NixOS/nixpkgs@7845b89](https://github.com/NixOS/nixpkgs/commit/7845b89e24ff9e91e6e13f66ee5a1a73958d22fb) address that?

Looks like this could fix the issue.

Mic92:
I no longer have the "busy mount" issue.

zowoq:
> > It doesn't migrate old mutable state into the writable overlay (might be not an issue here).
> 
> Yes, shouldn't be an issue here.

I forgot that we do have some mutable state that we need to handle, the ssh host key.

Mic92:
> > > It doesn't migrate old mutable state into the writable overlay (might be not an issue here).
> > 
> > 
> > Yes, shouldn't be an issue here.
> 
> I forgot that we do have some mutable state that we need to handle, the ssh host key.

At least with sops-nix it's possible to move to age keys outside of etc and deploy ssh keys with sops. This is what I tend to do these days.

zowoq:
Going to leave this for now.

PR #1442, opened 2024-09-09
Add doronbehar to trusted maintainers of both builders

doronbehar:
Would like to be able to test the Darwin builds of packages I maintain faster then ofborg. These packages include:

- `musescore`
- `scipy`
- `numpy`
- `openmpi`
- `gnuradio`

And more.


doronbehar:
Thanks for the quick response! When should I expect being able to ssh to the builder?

zowoq:
It was deployed automatically, you should be able to access both of them.

doronbehar:
Very impressive! Indeed I tested that I got access to the nixos build-box, however, I had a small passphrase mistake with the Darwin key, and I need to update it please :pray: https://github.com/nix-community/infra/pull/1443

PR #1466, opened 2024-09-22
modules/nixos: add cgroups and exporter

zowoq:


flake.nix
```diff
@@ -16,6 +16,8 @@
     buildbot-nix.inputs.nixpkgs.follows = "nixpkgs";
     buildbot-nix.inputs.treefmt-nix.follows = "treefmt-nix";
     buildbot-nix.url = "github:qowoz/buildbot-nix/more-skipped-build-workers";
+    cgroup-exporter.inputs.nixpkgs.follows = "nixpkgs";
+    cgroup-exporter.url = "github:arianvp/cgroup-exporter";
```
zowoq, line 20: Still need to send a PR upstream for aarch64-linux support.

Mic92, line 20: What is the difference to this plugin? https://github.com/influxdata/telegraf/tree/master/plugins/inputs/cgroup

zowoq: Oops, I went with the `cgroup-exporter` suggestion without checking if telegraf supported it.

Looks like the exporter collects almost everything and doesn't currently support any config, telegraf doesn't collect anything by default, requires config.

Mic92: Does it make sense to collect everything? We mainly do alert monitoring rather than looking at performance metrics.

zowoq: I tried the plugin and some metrics had permission issues (telegraf user?) or were broken, exporter seems to work okay (or least doesn't seem to be missing anything notable).

Probably don't want to collect everything but might be easier to start with the exporter and try switching to the plugin later once we have a better idea of what we do want to collect rather than needing to configure it from the start.

Mic92: okay.

zowoq: > We mainly do alert monitoring rather than looking at performance metrics.

Maybe I should bring back the grafana dashboard? I removed it as no one was using it but maybe it would more useful now? IIRC people on matrix were talking about checking the load on the build boxes, dashboard might be a bit better than people running top.

PR #1474, opened 2024-09-26
community-builder: allow users to set a shell, darwin: add packages

zowoq:


modules/darwin/common/packages.nix
```diff
@@ -0,0 +1,19 @@
+{ lib, pkgs, ... }:
+{
+  # srvos
+
+  environment.systemPackages = map lib.lowPrio [
```
jian-lin, line 5: Could `emacs` (or `emacs-nox`) be added?

zowoq: Sure, I've added it to systemPackages in modules/shared/community-builder.nix.

PR #1480, opened 2024-09-27
community-builder: add btop, htop to systemPackages

GaetanLepage:
For easier monitoring of system resources I propose to add [btop](https://github.com/aristocratos/btop) and `htop` to the community builders packages.

### DIFF
--- a/modules/shared/community-builder.nix
+++ b/modules/shared/community-builder.nix
@@ -18,6 +18,7 @@
 
     # useful for people that want to test stuff
     environment.systemPackages = [
+      pkgs.btop
       pkgs.emacs
       pkgs.fd
       pkgs.git


zowoq:
Don't need to add htop on linux, some packages are only in darwin because on linux they already included in the srvos modules we use.

Just need to add btop here:

https://github.com/nix-community/infra/blob/c7d7f1184d0c24fcd7891320843df5291f59f98c/modules/shared/community-builder.nix#L20-L30

GaetanLepage:
> Don't need to add htop on linux, some packages are only in darwin because on linux they already included in the srvos modules we use.

Are you sure about this ?
I tried before adding it in:
```
[glepage@aarch64:~]$ htop
-bash: htop: command not found
```

zowoq:
Wrong machine? Looks like that is https://github.com/NixOS/aarch64-build-box?

GaetanLepage:
Oh right. I removed it.

PR #1488, opened 2024-09-30
modules/nixos: add nginx

zowoq:
~~Maybe quic as well?~~ Will look at quic in another PR.

~~`net.ipv4.tcp_fastopen=3` ?~~

### DIFF
--- a/modules/nixos/nginx.nix
+++ b/modules/nixos/nginx.nix
@@ -1,0 +1,53 @@
+{
+  config,
+  inputs,
+  lib,
+  pkgs,
+  ...
+}:
+{
+  options.services.nginx.virtualHosts = lib.mkOption {
+    type = lib.types.attrsOf (
+      lib.types.submodule {
+        config = {
+          enableACME = lib.mkDefault true;
+          forceSSL = lib.mkDefault true;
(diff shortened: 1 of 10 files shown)


modules/nixos/nginx.nix
```diff
@@ -0,0 +1,53 @@
...
+          '';
+
+          locations."= /robots.txt".alias = pkgs.writeText "robots.txt" ''
+            User-agent: *
+            Disallow: /
+          '';
+        };
+      }
+    );
+  };
```
zimbatm, line 28: I didn't know you could re-define existing options

Mic92: Submodules are also modules at the end of the day.

Mic92:
Did you measure the performance of KTLS? I measured it for harmonia and it actually made performance measurable worse.

zowoq:
Sorry, didn't realise that you posted two comments, when I opened the notification I only saw the last one.

No, I didn't check this specific configuration but about a year ago I did a similar nginx config for another project and it seemed fine. Can we try the same measurements here? How different is the config you are using, e.g. worker_processes, quic, reuseport?

PR #1507, opened 2024-10-12
set gcc.arch, extra-platforms, system-features

zowoq:


hosts/build01/default.nix
```diff
@@ -1,12 +1,15 @@
...
   imports = [
-    inputs.srvos.nixosModules.hardware-hetzner-online-amd
-    inputs.self.nixosModules.disko-zfs
-    inputs.self.nixosModules.builder
     inputs.self.nixosModules.community-builder
+    inputs.self.nixosModules.disko-zfs
+    inputs.srvos.nixosModules.hardware-hetzner-online-amd
   ];
 
+  nix.settings.max-jobs = 24;
```
Mic92, line 9: Is this the same as number of cores for each machine?

zowoq: Yes.

PR #1514, opened 2024-10-16
tasks.py: various

zowoq:


Mic92:
What is the advantage of passing hosts in explicitly?

zowoq:
> Does this not just make it less convenient to test a change on all machines.

I don't test changes on all machines simultaneously.

zowoq:
Also it is slow to eval and deploying to web02 is slow anyway, I'd rather pass them explicitly and get an error if not.

PR #1515, opened 2024-10-16
terraform/hydra-nixpkgs: add unfree-redist-darwin

zowoq:
I'll experiment with this later after `unfree-redist` is merged, not sure if we really have the cycles to spare for it with only the one darwin CI builder.

https://hydra.nix-community.org/jobset/nixpkgs/unfree-redist-darwin


zowoq:
@zimbatm If you want to have an unfree darwin jobset it seems feasible, I've watched a couple of full rebuilds, takes 4-5 hours with hydra limited to a single job so we reverse some darwin capacity for other CI jobs.

zimbatm:
nice!

PR #1528, opened 2024-10-22
docs: add infrastructure

zowoq:
I removed comments from the hosts doc in ff60d32421cf8389ebf358477fd5ec4cfbfc2148 as didn't need them for an internal doc and didn't want to have to keep updating them. 

No reason we can't make this doc more visible and we can link to the service pages instead of adding comments.


zimbatm (approved):
good idea

Mic92 (approved):

PR #1530, opened 2024-10-22
terraform/cloudflare: refactor

zowoq:

### DIFF
--- a/terraform/cloudflare.tf
+++ b/terraform/cloudflare.tf
@@ -4,7 +4,10 @@ locals {
     for role in data.cloudflare_account_roles.account_roles.roles :
     role.name => role
   }
-  cf_admins = local.admins
+  cf_admins = {
+    for key, value in local.admins :
+    key => value if key != "adisbladis"
+  }
 }
 
 data "cloudflare_account_roles" "account_roles" {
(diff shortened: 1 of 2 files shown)


zimbatm:
Yes, it looks functionally correct. I'm not sure what prompted the move. Was the code getting in the way of something else?

zowoq:
Now that we're going to be advertising the hydra jobsets I thought it could do with a refactor, it is verbose but not very readable.

Mic92:
From my personal experience a little bit more verbosity in terraform is better than too much abstraction. Infrastructure automation is just different from programming because it also involves state management. i.e. for targeted applies its nice if one can copy'n'paste target names. I am sure this is still possible, but it just requires more effort and reading now. But feel free to follow whichever branch you want.

terraform/cloudflare.tf
```diff
@@ -4,7 +4,10 @@ locals {
     for role in data.cloudflare_account_roles.account_roles.roles :
     role.name => role
   }
-  cf_admins = local.admins
+  cf_admins = {
+    for key, value in local.admins :
+    key => value if key != "adisbladis"
```
zowoq, line 9: Skipping this invite as we keep resending it.

PR #1533, opened 2024-10-24
update nix-darwin, remove users.forceRecreate

zowoq:


modules/darwin/community-builder/users.nix
```diff
@@ -357,7 +357,5 @@ in
 
   users.knownUsers = builtins.map (u: u.name) users;
 
-  users.forceRecreate = true;
```
zowoq, line 360: Since the upstream change from dscl to sysadminctl this option now also deletes the users $HOME>

PR #1536, opened 2024-10-26
nixpkgs-update: add queue

zowoq:
> it is technically possible for you to download the supervisor SQLite state database from https://nixpkgs-update-logs.nix-community.org/~supervisor/state.db and issue the appropriate queries against it (after having studied https://github.com/nix-community/infra/blob/master/hosts/build02/supervisor.py to understand the queue schema). But it's not exactly end-user friendly at the moment. (Some day I may publish a dashboard.)

@rhendric I thought I'd throw together something basic so people don't need to play around with the database. I'm fine with replacing this with your dashboard, if you do make one.

https://nixpkgs-update-logs.nix-community.org/~supervisor/queue.html


rhendric:
Good idea!

The query I wrote a while back to view the queue is more complicated:

```sql
select attr_path, last_started, finished
from queue
left join log using (attr_path)
group by last_started, order_jitter, attr_path
having is_dequeued = 0
and (started is null or finished is not null or started + 43200 < $now)
and (true or max(fetcher_run_started))
order by last_started asc, order_jitter, attr_path;
```

I don't remember verifying how much of that extra complexity is actually necessary, and I should. Poke me if I don't come back to this in a few days.

zowoq:
> Poke me if I don't come back to this in a few days.

@rhendric poke.

Also, I'm not sure what you had in mind for your dashboard but I've deployed a grafana instance for the community infra, would that be of any use?

PR #1538, opened 2024-10-27
terraform/hydra-projects: add microvm.nix

astro:
I saw the NixCon lightning talk yesterday and it reminded me that I'd really love to have some test runners for microvm.nix.

I don't want to move the repo to this org though for stability and vanity reasons. Is that okay?

### DIFF
--- a/terraform/hydra-projects.tf
+++ b/terraform/hydra-projects.tf
@@ -69,3 +69,30 @@ resource "hydra_project" "simple_nixos_mailserver" {
     value = "https://gitlab.com/simple-nixos-mailserver/nixos-mailserver"
   }
 }
+
+resource "hydra_project" "microvm_nix" {
+  name         = "microvm-nix"
+  display_name = "MicroVM.nix"
+  description  = "NixOS MicroVMs"
+  homepage     = "https://github.com/astro/microvm.nix"
+  owner        = "admin"
+  enabled      = true
+  visible      = true
+}
+
+resource "hydra_jobset" "microvm_nix" {
+  project     = hydra_project.microvm_nix.name
+  state       = "enabled"
+  visible     = true
+  name        = "main"
+  type        = "flake"
+  description = "main branch"
+
+  flake_uri = "github:astro/microvm.nix"
+
+  check_interval    = 1800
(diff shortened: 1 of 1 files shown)


zowoq:
> test runners for microvm.nix.

> I don't want to move the repo to this org though

This PR would only give you access to our public build box (12 core x86_64-linux).

We don't give out ssh access to our 48 core x86_64-linux, 80 core aarch64-linux machines. If you want access to our buildbot CI (for PR tests) the repo would need to be moved, otherwise we could give you a jobset on our hydra?

astro:
A jobset on Hydra would be sufficient. The flakeref is `github:astro/microvm.nix`

zowoq:
https://hydra.nix-community.org/jobset/microvm-nix/main

astro:
Thank you very much!

zowoq:
@astro My fault for not checking, I was expecting a flake with some VM tests (e.g. ~20) but it seems you have ~200. 

Unfortunately our hydra is configured to best work with the cuda jobset, which means only 2 concurrent jobs on each machine, the amount of VM tests doesn't work well with that as some of them seem to have long timeouts or otherwise get stuck, blocking other jobs. I bumped concurrent jobs and restarted to try to get one one complete eval but I still ended up cancelling several of the vm tests.

The concurrent jobs problem is something is we're working on, I'm hoping that we'll make progress on that in the next few weeks but for now I've disabled the microvm jobset so it doesn't interfere with our existing jobsets.

zimbatm:
The https://github.com/apps/buildbot-nix-community app could be installed on the repo instead.

zowoq:
> The https://github.com/apps/buildbot-nix-community app could be installed on the repo instead.

How can we run it as a public app without it being a security issue?

Mic92:
We could add an org whitelist feature to buildbot-nix.

zowoq:
> We could add an org whitelist feature to buildbot-nix.

That'll work, though I'd suggest scoping it to repos.

There is also https://github.com/nix-community/buildbot-nix/issues/176 which looks like it would be very similar to a hydra jobset.

PR #1542, opened 2024-10-29
modules/shared/community-builder: add github token to nix.conf

zowoq:


secrets/secrets.nix
```diff
@@ -11,13 +11,20 @@ let
...
+  build01 = knownHosts.build01.publicKey;
   build02 = knownHosts.build02.publicKey;
   build03 = knownHosts.build03.publicKey;
   build04 = knownHosts.build04.publicKey;
+  darwin01 = knownHosts.darwin01.publicKey;
   darwin02 = knownHosts.darwin02.publicKey;
   web02 = knownHosts.web02.publicKey;
 
   secrets = {
+    # fine-grained, no permissions github token from `nix-infra-bot`, expires 2025-10-29
```
Mic92, line 23: Has the bot still any permission in our infrastructure?
Otherwise I would suggest to make this a new bot, just in case github messes up their API at some point in the future.

Mic92: I can also make one if this is the limiting factor.

zowoq: > Has the bot still any permission in our infrastructure?

No. Currently it is used in two places, a collaborator in the nixvim repo for mergify to use for rebasing PRs and another token without permissions is used for the nixpkgs-update-github-releases service.

zowoq: We also still have the `nix-community-buildbot` user account that isn't used for anything at all, I'll use that instead.

PR #1548, opened 2024-11-03
modules/nixos/monitoring: import hosts from knownHosts

zowoq:
Rather than using another list of hosts we can reuse known hosts.

### DIFF
--- a/modules/nixos/monitoring/prometheus.nix
+++ b/modules/nixos/monitoring/prometheus.nix
@@ -1,4 +1,4 @@
-{ pkgs, ... }:
+{ inputs, pkgs, ... }:
 {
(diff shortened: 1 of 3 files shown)


modules/nixos/monitoring/prometheus.nix
```diff
@@ -19,11 +19,13 @@
         metrics_path = "/metrics";
         static_configs =
           let
-            hosts = import ./hosts.nix;
+            hosts = (import "${inputs.self}/modules/shared/known-hosts.nix").programs.ssh.knownHosts;
```
Mic92, line 22: Using something like self.nixosModules.known-hosts was not possible?

Mic92 (approved):

PR #1552, opened 2024-11-04
docs/community-builder: add example NixOS configuration for remote builds

JohnRTitor:
Added example NixOS configuration for remote builds using `nix-build` and `nix build` commands.

Inspired from https://github.com/NixOS/aarch64-build-box?tab=readme-ov-file#configuring-your-computer-for-remote-builds

[Tested in my configuration](https://github.com/JohnRTitor/nix-conf/commit/7f4a74c1dfb5d074306305a2c76e58a76c7af7ab), now finally nixpkgs-review works!

### DIFF
--- a/docs/community-builder.md
+++ b/docs/community-builder.md
@@ -57,3 +57,77 @@ $ $path
 ```
 
 _(My [implementation](https://github.com/ckiee/nixfiles/blob/aac57f56e417e31f00fd495d8a30fb399ecbc19b/deploy/hm-only.nix#L10) of [this](https://github.com/ckiee/nixfiles/blob/aac57f56e417e31f00fd495d8a30fb399ecbc19b/bin/c#L92-L95) ~ckie)_
+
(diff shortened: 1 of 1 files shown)


zowoq:
> Added example NixOS configuration for remote builds using `nix-build` and `nix build` commands.
> 
> Inspired from https://github.com/NixOS/aarch64-build-box?tab=readme-ov-file#configuring-your-computer-for-remote-builds

> Added a warning after discussing in the matrix channel.

I intentionally didn't add this, didn't want to have an "example" `distributedBuilds` config, even with a warning.

> now finally nixpkgs-review works!

Maybe we could show an example of passing `builders` to `nixpkgs-review` instead.

JohnRTitor:
> Maybe we could show an example of passing `builders` to `nixpkgs-review` instead.

I am not sure how it works actually. 

But I thought it would be safe if we configure it where NixOS system platform != Builder platform

PR #1561, opened 2024-11-11
flake: expose {darwin,nixos}Modules automatically

zowoq:
@zimbatm Do you mind if we reuse this code from `blueprint`?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -60,6 +60,7 @@
       systems = import inputs.systems;
 
       imports = [
+        ./modules
         inputs.lite-config.flakeModule
         inputs.treefmt-nix.flakeModule
       ];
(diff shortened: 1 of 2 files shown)


zimbatm:
I don't mind, but I'm curious: how would blueprint have to change to be useful for the infra repo?

zowoq:
I don't think there is anything specific that would need to change in `blueprint`, I think what we'd need to consider is do we want to opt out of the `flake-parts`/`flakeModule` ecosystem, e.g. effects are useable without flake-parts but the integration is nicer.

zowoq:
I did a first pass on blueprint in https://github.com/nix-community/infra/pull/1563. I'll probably continue working on it occasionally but I'd likely postpone any decision on switching until after we've using effects on buildbot.

PR #1562, opened 2024-11-11
modules/nixos/common/update: switch to nixos-rebuild --store-path

zowoq:
nixos-rebuild using systemd-run suppresses the output of pre-switch checks so without the patch we lose the update diff.

I'll see about upstreaming the patch in a week or two.

### DIFF
--- a/modules/nixos/common/update.bash
+++ b/modules/nixos/common/update.bash
@@ -2,6 +2,8 @@ arch=$(uname -m)
 hostname=$(uname -n)
 p=$(curl -L https://buildbot.nix-community.org/nix-outputs/nix-community/infra/master/"$arch"-linux.host-"$hostname")
 
+export NIXOS_REBUILD_NO_SYSTEMD_RUN=1
+
 cancel_reboot() {
   if [[ -e /run/systemd/shutdown/scheduled ]]; then
     shutdown -c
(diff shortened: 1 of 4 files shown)


modules/nixos/common/update.bash
```diff
@@ -6,25 +6,36 @@ if [[ "$(readlink /run/current-system)" == "$p" ]]; then
   exit 0
 fi
 
-nix-store --option narinfo-cache-negative-ttl 0 --realise "$p"
-nix-env --profile /nix/var/nix/profiles/system --set "$p"
+nix-store \
```
Mic92, line 9: I would like to add this code eventually to buildbot-nix in its current form (also with support for both macos & NixOS).

zowoq: macos support is complicated, nix-darwin doesn't have something like `restartIfChanged` for services and there isn't a way of monitoring failed services like we have on linux. I was planning to just keep using push deployments for darwin but switching from actions to effects.

zimbatm: Would it be possible to come up with a common interface for both? Otherwise deploy tools all have to take care of those differences.

zowoq: Sorry, I don't understand what a "common interface" means here if it isn't a deployment tool?

PR #1563, opened 2024-11-12
WIP blueprint

zowoq:


flake.nix
```diff
@@ -54,101 +56,51 @@
...
-              })
-            ];
-          };
+  # overlays = [
+  #   (final: prev: {
+  #     hydra = (prev.hydra.override { nix = final.nixVersions.nix_2_24; }).overrideAttrs (o: {
+  #       version = inputs.hydra.shortRev;
+  #       src = inputs.hydra;
+  #       buildInputs = o.buildInputs ++ [ final.perlPackages.DBIxClassHelpers ];
+  #     });
```
zimbatm, line 65: This could be moved to `packages/hydra/default.nix`, and then pulled in the module with `services.hydra.package = perSystem.self.hydra`;

flake.nix
```diff
@@ -54,101 +56,51 @@
...
+  #   })
+  # ];
 
-          hostModuleDir = ./hosts;
+  # treefmt = {
+  #   flakeCheck = system == "x86_64-linux";
+  #   imports = [ ./dev/treefmt.nix ];
+  # };
```
zimbatm, line 72: You can create a `package/formatter/default.nix` for this.

See https://github.com/zimbatm/home/blob/main/packages/formatter/default.nix

flake.nix
zowoq: With lite-config this overlay is applied to `pkgs`so we don't need to do any additional wiring, e.g. `.#nixosConfigurations.build03` and `.#checks.x86_64-linux.nixosTests-hydra` both just work. It is also exposed directly in `.#packages.${system}.hydra`.

zowoq:
Autoloading host config, modules, devshells, packages is good but I think I'd rather just handle the checks manually. Some things doesn't need to be exposed except as a check and others I want exposed on all platforms but CI should only build one platform.

___

Minor nit, might be nice to hide these if they aren't used but maybe that is something better handled by flake schemas. 
```
❯ nix flake show

homeModules: unknown
lib: unknown
modules: unknown
templates
```

zimbatm:
Thanks for the feedback. For the checks, would it help if I added a checks/ folder that works like the packages/ folder? To restrict the checks to specific platforms, you can annotate them with the `meta.platforms` attibute.

zowoq:
> For the checks, would it help if I added a checks/ folder that works like the packages/ folder? To restrict the checks to specific platforms, you can annotate them with the meta.platforms attibute.

No, I realise that follows the existing pattern but checks folder and meta.platforms but it makes it more complicated than it needs to be.

This is fine, just need a way to do it.

```
checks =
  let
    darwinConfigurations = lib.mapAttrs' (
      name: config: lib.nameValuePair "host-${name}" config.config.system.build.toplevel
    ) ((lib.filterAttrs (_: config: config.pkgs.system == system)) self.darwinConfigurations);
    devShells = lib.mapAttrs' (n: lib.nameValuePair "devShell-${n}") self'.devShells;
    nixosConfigurations = lib.mapAttrs' (
      name: config: lib.nameValuePair "host-${name}" config.config.system.build.toplevel
    ) ((lib.filterAttrs (_: config: config.pkgs.system == system)) self.nixosConfigurations);
  in
  darwinConfigurations
  // devShells
  // {
    inherit (self') formatter;
  }
  // nixosConfigurations
  // pkgs.lib.optionalAttrs (system == "x86_64-linux") {
    inherit (self'.packages) docs docs-linkcheck;
    nixpkgs-update-supervisor-test = pkgs.callPackage ./hosts/build02/supervisor_test.nix { };
    nixosTests-buildbot = pkgs.nixosTests.buildbot;
    nixosTests-buildbot-nix-master = inputs'.buildbot-nix.checks.master;
    nixosTests-buildbot-nix-worker = inputs'.buildbot-nix.checks.worker;
    nixosTests-hydra = pkgs.nixosTests.hydra.hydra;
  };
```

zowoq:
> What if you could extend the checks more easily, and let the rest be managed by blueprint?

Isn't this going to end up being basically same as what I was asking for some of the default ones will need to be disabled?

zowoq:
I rebased this on master to take another look at it but I think prefer the status quo.

zimbatm:
Sounds good. Is there something in particular that tilts the balance?

zowoq:
No, nothing specific. If it was reversed and we had been using blueprint for the last couple of years and the suggestion was to move to flake-parts I think I'd say the same thing.

I'm not actually planning to do this but at the moment if we were to make a change I'd probably try switching back to the old style of pinning with npins or similar.

PR #1566, opened 2024-11-14
build05: init (aarch64-build-box.nix-community.org)

zowoq:
Hetzner RX170 https://www.hetzner.com/dedicated-rootserver/rx170

This is to replace https://github.com/NixOS/aarch64-build-box when the Equinix Metal sponsorship ends.

Fundraising target is at https://opencollective.com/nix-community.

I'll handle the armv7l kernel personality patch that the current machine has in https://github.com/nix-community/infra/pull/1325.


Mic92:
Can we make that oracle box, a community builder. So we have something for arm at least?

zowoq:
On our existing aarch64 hetzner machine we could set up a nixos container, limit it to 50% cpu/mem and allow direct ssh access, might be better than the oracle vm?

When does the equinix sponsorship end?

Mic92:
> On our existing aarch64 hetzner machine we could set up a nixos container, limit it to 50% cpu/mem and allow direct ssh access, might be better than the oracle vm?
> 
> When does the equinix sponsorship end?

End of the year.

PR #1567, opened 2024-11-15
modules/nixos/monitoring: remove ofborg

zowoq:


modules/nixos/monitoring/telegraf.nix
```diff
@@ -40,6 +40,5 @@
         tags.org = "nix-community";
         timeout = "10s";
       }) (builtins.attrValues hosts);
-    prometheus.urls = [ "https://events.ofborg.org/prometheus.php" ];
```
Mic92, line 43: We might get it back, but I don't think this php will be there.

PR #1580, opened 2024-11-22
modules/darwin/community-builder: add user artemist

artemist:
I am working on better support for BSDs in nixpkgs. While this doesn't intentionally affect darwin, sometimes I cause mass rebuilds and would like to be able to test on an `aarch64-darwin` machine.
I already have `x86_64-linux` machines so I do not need access to the linux builder.


zimbatm:
Would it be helpful to have access to a FreeBSD machine? We could consider making one available.

artemist:
Unfortunately we haven't fully upstreamed FreeBSD sandboxing to cppnix or lix yet so it's a bit premature for a nix-community FreeBSD builder. I have a fairly powerful x86_64-freebsd machine at home and would be willing to help maintain a community FreeBSD machine once lix is in a good state.

zimbatm:
Sounds great.

I don't know if sandboxing is a blocker. Even with the sandboxing, local exploits are common.

Our primary defence is that our community members are decent people (luckily :sweat_smile:). And the build boxes don't hold any secrets.

artemist:
Yeah, that's fair. The sandboxing is less for security here and more for purer builds, since we don't want to mysteriously use the host libc or something

zowoq:
As there is some prior discussion about bsd I'll mention it here.

@rhelmot You mentioned native CI on a nixbsd issue a few hours ago, nix-community could probably help with that. I've already been taking a look at setting up nixbsd VMs as builders on our existing infra.

These are our current machines: https://nix-community.org/infrastructure

Something like this to start with:
```
- x86_64-linux build box
  - freebsd VM

- x86_64-linux CI builder
  - freebsd VM
```

If you wanted aarch64 or openbsd on either arch we could probably do that as well.

rhelmot:
x86_64-freebsd is definitely the priority, let's start with that. If you got that set up for me, I would be so happy I could cry.

artemist:
I guess this is an excuse for me to rebase my lix patches

zowoq:
Working on builder VMs in https://github.com/nix-community/infra/pull/1734.

I've also created a hydra jobset for the current x86_64-linux cross builds.

PR #1595, opened 2024-12-04
build01,build02: enable cgroups, modules/nixos/common: increase zram to 100%

zowoq:


modules/nixos/common/default.nix
```diff
@@ -33,6 +33,7 @@
   ) pkgs.linuxPackages_latest;
 
   zramSwap.enable = true;
+  zramSwap.memoryPercent = 100;
```
zowoq, line 36: @Mic92 Default is 50, I don't think there is any problem with increasing this on all hosts?

PR #1599, opened 2024-12-07
build01: switch to zfs_unstable

zowoq:
@Mic92 Do we need to make any changes for Direct I/O, looks like it is enabled by default?


Mic92:
No direct I/O should just work.

PR #1630, opened 2024-12-28
modules/nixos/common/builder: enable kernel samepage merging

zowoq:
This is enabled in nixos/infra.

https://docs.kernel.org/admin-guide/mm/ksm.html


Mic92:
Maybe useful with VM tests, but I don't think we enable this in qemu?

zowoq:
> Maybe useful with VM tests

Yes, I'm hoping it'll save a little bit of memory when vm tests are running in parallel.

> but I don't think we enable this in our test framework

I don't understand why it would need to be?

PR #1635, opened 2025-01-02
nixpkgs-update: filter beam package set to a single set

adamcstephens:
There are now multiple beam package sets, so let's limit to updating one.


adamcstephens:
Pinging @zowoq before we get any more bumps for multiple versions.

zowoq:
What about `beamMinimal*Packages`?

adamcstephens:
Good call. Updated.

adamcstephens:
Thanks!

PR #1638, opened 2025-01-03
modules/nixos/common: use preSwitchChecks for update diff

zowoq:
Second attempt at adding a boot diff (https://github.com/nix-community/infra/pull/1609) but using preSwitchChecks (https://github.com/NixOS/nixpkgs/commit/6e192c4489ea44c48876faa11060798fb9502146) which works for boot/switch, grub/systemd-boot and has an escape hatch which can be used for things like disko install tests, nixos-anywhere, etc.

I've move this to srvos after we've tested it here for a bit.

### DIFF
--- a/modules/nixos/common/default.nix
+++ b/modules/nixos/common/default.nix
@@ -17,6 +17,15 @@
     inputs.srvos.nixosModules.server
   ];
 
+  srvos.update-diff.enable = false;
+  system.preSwitchChecks.update-diff = ''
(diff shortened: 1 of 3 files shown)


zimbatm:
No objections but what was the issue with the current approach from `srvos.update-diff.enable`?

zowoq:
The current approach in srvos only works for `switch-to-configuration switch`.

PR #1640, opened 2025-01-05
tasks.py: use copy instead of archive for deploy

zowoq:
this is faster but it assumes that the flake inputs can be substituted on the target


Mic92:
Also use this most of the time. Just for path inputs one has to use archive.

PR #1656, opened 2025-01-09
nixos/community-builder: add user zimward

zimward:
Having access to an `aarch64` machine that will not OOM on `nixpkgs-review` would be very helpful. (And not having to wait a day to test a change for https://github.com/NixOS/nixos-hardware/pull/1073)


zowoq:
Apologies, for not following up on this sooner.

We don't really have specific rules for allowing access to this machine but it is basically limited to people who are well known in the nixpkgs/nixos/etc community, is there somebody who is known (e.g. nixpkgs committer) who can vouch for you?

zimward:
I interacted a bit with Atemu on Discourse and on my PR on `FHSenv` overriding, i also talked with lassulus once at c-base (though i don't think that should give me any credibility except confirming that i am in fact a real person).

Lassulus:
I talked to @zimward today again. They seem cool. Would trust them to not do any shenanigans with the builder :)

PR #1659, opened 2025-01-11
nixos/community-builder: add wolfgangwalther

wolfgangwalther:
Already have access to the darwin builder, would like to build on aarch64-linux with nixpkgs-review, too.

The file says "alphabetical order" at the top (although that comment seems to be outdated), but clearly the list is not sorted anymore. Luckily... I will go at the end in any case!

Thanks!


zowoq:
> The file says "alphabetical order" at the top (although that comment seems to be outdated), but clearly the list is not sorted anymore. 

Updating the comment in https://github.com/nix-community/infra/pull/1663.

PR #1667, opened 2025-01-15
nixos/community-builder: add pbsds

pbsds:
I used to have access to NixOS/aarch64-build-box, and it would be nice to be able to build some aarch64 again without binfmt emulation


modules/nixos/community-builder/users.nix
```diff
@@ -193,6 +193,12 @@ let
       trusted = true;
       keys = ./keys/winter;
     }
+    {
+      # lib.maintainers.pbsds, https://github.com/pbsds
+      name = "pbsds";
+      trusted = true;
+      keys = ./keys/pbsds;
+    };
```
zimbatm, line 201:
```suggestion
    }
```

PR #1675, opened 2025-01-19
tasks.py: use archive for deploy when the flake has file or path inputs

zowoq:
Switched to copy in https://github.com/nix-community/infra/pull/1640 but archive it still useful sometimes.


Mic92 (approved):
Looks similar to what I used before.

PR #1678, opened 2025-01-21
modules/nixos/common: reboot on panic

zowoq:
Try to avoid having stuck machines that need to have their power cycled manually.

### DIFF
--- a/modules/nixos/common/default.nix
+++ b/modules/nixos/common/default.nix
@@ -35,6 +35,26 @@
   zramSwap.enable = true;
   zramSwap.memoryPercent = 100;
 
+  systemd.services.sysctl-after-boot = {
+    serviceConfig.Restart = "on-failure";
+    serviceConfig.Type = "oneshot";
+    path = [
+      pkgs.procps
+    ];
+    script = ''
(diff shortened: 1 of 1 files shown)


modules/nixos/common/default.nix
```diff
@@ -35,9 +35,15 @@
   zramSwap.enable = true;
   zramSwap.memoryPercent = 100;
 
-  # https://github.com/NixOS/nixpkgs/pull/268121
-  # https://wiki.archlinux.org/title/Zram#Optimizing_swap_on_zram
   boot.kernel.sysctl = {
+    "kernel.hardlockup_panic" = 1;
+    "kernel.hung_task_panic" = 1;
+    "kernel.panic_on_oops" = 1;
```
Mic92, line 41: When are those applied in the boot process? I would probably apply those late in the boot process (maybe even 5mins after boot) as it can potentially make it hard to debug early boot issues i.e. faulty hardware + kvm console. Those issues would also likely cause a reboot loop anyway.

zowoq: Good point, moved to run 5m after boot.

PR #1689, opened 2025-01-29
build01: test zram writebackDevice

zowoq:

### DIFF
--- a/hosts/build01/default.nix
+++ b/hosts/build01/default.nix
@@ -7,6 +7,9 @@
     inputs.srvos.nixosModules.hardware-hetzner-online-amd
   ];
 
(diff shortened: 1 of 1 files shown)


hosts/build01/default.nix
```diff
@@ -7,6 +7,9 @@
     inputs.srvos.nixosModules.hardware-hetzner-online-amd
   ];
 
+  # zfs create zroot/writeback -V 128G
+  zramSwap.writebackDevice = "/dev/zvol/zroot/writeback";
```
Mic92, line 11: Hopefully this doesn't deadlock :)
Is this known to work with zvol?

PR #1598, opened 2024-12-07
modules/nixos/buildbot: add microvm.nix

zowoq:
Didn't start automatically so I triggered an eval manually:
```
warning: ignoring the client-specified setting 'trusted-public-keys', because it is a restricted setting and you are not a trusted user
warning: ignoring untrusted substituter 'https://microvm.cachix.org', you are not a trusted user.
Run `man nix.conf` for more information on the `substituters` configuration option.
warning: ignoring the client-specified setting 'trusted-public-keys', because it is a restricted setting and you are not a trusted user
error: flake 'git+file:///var/lib/buildbot-worker/worker-081/astro_microvm_nix_nix-eval/build' does not provide attribute 'checks'
warning: ignoring untrusted substituter 'https://microvm.cachix.org', you are not a trusted user.
Run `man nix.conf` for more information on the `substituters` configuration option.
warning: ignoring the client-specified setting 'trusted-public-keys', because it is a restricted setting and you are not a trusted user
error: flake 'git+file:///var/lib/buildbot-worker/worker-081/astro_microvm_nix_nix-eval/build' does not provide attribute 'checks'
warning: ignoring untrusted substituter 'https://microvm.cachix.org', you are not a trusted user.
Run `man nix.conf` for more information on the `substituters` configuration option.
warning: ignoring the client-specified setting 'trusted-public-keys', because it is a restricted setting and you are not a trusted user
error: flake 'git+file:///var/lib/buildbot-worker/worker-081/astro_microvm_nix_nix-eval/build' does not provide attribute 'checks'
warning: ignoring untrusted substituter 'https://microvm.cachix.org', you are not a trusted user.
Run `man nix.conf` for more information on the `substituters` configuration option.
warning: ignoring the client-specified setting 'trusted-public-keys', because it is a restricted setting and you are not a trusted user
error: flake 'git+file:///var/lib/buildbot-worker/worker-081/astro_microvm_nix_nix-eval/build' does not provide attribute 'checks'
error: flake 'git+file:///var/lib/buildbot-worker/worker-081/astro_microvm_nix_nix-eval/build' does not provide attribute 'checks'
warning: ignoring untrusted substituter 'https://microvm.cachix.org', you are not a trusted user.
Run `man nix.conf` for more information on the `substituters` configuration option.
warning: ignoring the client-specified setting 'trusted-public-keys', because it is a restricted setting and you are not a trusted user
warning: ignoring untrusted substituter 'https://microvm.cachix.org', you are not a trusted user.
[...]

PR #1707, opened 2025-02-14
modules/nixos/buildbot: disable topic, switch to repoAllowlist

zowoq:

### DIFF
--- a/docs/continuous-integration.md
+++ b/docs/continuous-integration.md
@@ -17,9 +17,7 @@ See [here](./infrastructure.md#continuous-integration) for details about the har
 
 _Buildbot is the only CI system that supports pull requests from forked repositories._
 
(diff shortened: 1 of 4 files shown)


zowoq:
> https://github.com/nix-community/infra/pull/1598#discussion_r1957073931
> Isn't the github topic less work overall than having this list synchronously managed?

Responding here as the other PR was rebased on this one.

For users in nix-community, yes, it is more work as now they either need to send a PR or ask me to do it.

For users outside nix-community, no, as they only need to add the repo to the list, they don't need to bother with adding the topic.

For admins, no, the list is less work and simpler. Buildbot usually needs to be restarted manually anyway to pick up topic changes and it also allows me to enable/disable repos temporarily or permanently without needing to make changes to the repos, adding/removing topics or unarchiving/archiving repos.

We could revisit this in the future but at the moment this is preferable.

PR #1714, opened 2025-02-24
flake: switch to shallow nixpkgs clone

zowoq:

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -33,7 +33,7 @@
     nixpkgs-update.inputs.mmdoc.follows = "empty";
     nixpkgs-update.inputs.treefmt-nix.follows = "treefmt-nix";
     nixpkgs-update.url = "github:nix-community/nixpkgs-update/infra";
-    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable-small";
+    nixpkgs.url = "git+https://github.com/NixOS/nixpkgs?shallow=1&ref=nixos-unstable-small";
     nur-update.inputs.nixpkgs.follows = "nixpkgs";
     nur-update.url = "github:nix-community/nur-update";
     sops-nix.inputs.nixpkgs.follows = "nixpkgs";
--- a/flake.lock
+++ b/flake.lock
@@ -196,18 +196,19 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1746554066,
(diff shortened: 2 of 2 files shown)


flake.nix
```diff
@@ -36,7 +36,7 @@
     nixpkgs-update.inputs.mmdoc.follows = "empty";
     nixpkgs-update.inputs.treefmt-nix.follows = "treefmt-nix";
     nixpkgs-update.url = "github:nix-community/nixpkgs-update/infra";
-    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable-small";
+    nixpkgs.url = "https://nixos.org/channels/nixos-unstable-small/nixexprs.tar.xz";
```
Mic92, line 39: Actually this url is faster if you regularly switch between nixpkgs versions:

```
--- flake.nix
+++ flake.nix
@@ -2,7 +2,7 @@
   description = "A universal nixos installer, just needs ssh access to the target system";
 
   inputs = {
-    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable-small";
+    nixpkgs.url = "git+https://github.com/NixOS/nixpkgs?shallow=1&ref=nixos-unstable-small";
     flake-parts = { url = "github:hercules-ci/flake-parts"; inputs.nixpkgs-lib.follows = "nixpkgs"; };
 
     # used for testing
```

That cache should work both locally and in buildbot. For fresh clones it's almost as fast as downloading the github archive (tested in github actions)

zowoq: IIRC github has asked projects (homebrew and others) to stop using shallow clones because they are expensive to run on the servers.

Mic92: It's only a shallow clone on the first run because than the cache takes over and it's just a handful commit that is fetched (which should be cheaper than the github archive).
Especially if your internet is not great.

zowoq: That isn't what I was referring to, just using it in this repo isn't the issue, other people copying the pattern is.

zowoq:
`unpacking 'https://releases.nixos.org/nixos/unstable-small/***/nixexprs.tar.xz' into the Git cache...` takes a couple of minutes, not sure why.

flake.nix
zowoq: I see the plan is for nix to migrate from `github:` to shallow `git:` so we may as well switch to your suggestion.

PR #1734, opened 2025-03-17
nixbsd freebsd builder

zowoq:
- ~~currently can't build anything in the vm~~:
  -  ~~`
error: build of '/nix/store/hicd8c5cnfnzc91kjqfhwdrc2d99n9mq-bootstrap-archive.drv' on 'ssh-ng://nix@freebsd-builder' failed: error: Failed to mount nullfs for /nix/store/hicd8c5cnfnzc91kjqfhwdrc2d99n9mq-bootstrap-archive.drv.chroot/bin/sh - Invalid fstype: Invalid argument
`~~
  - ~~`error: builder for '/nix/store/*.drv' failed due to signal 11 (Segmentation fault)`~~
- ~~initialHashedPassword~~
  - Won't bother with this or putting ssh keys in sops for now. Neither of the VMs are directly exposed, on build01 (build box) it doesn't really matter if we allow everyone with ssh access to the machine to also have root in the VM and on build03 (CI) only the community admins have ssh access to the machine anyway.
- ~~telegraf: `ld-elf.so.1: Shared object "libthr.so.3" not found, required by "telegraf"`~~
- ~~increase disk size~~

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -137,6 +137,62 @@
         "type": "github"
       }
     },
+    "lix": {
+      "inputs": {
+        "flake-compat": [
+          "nixbsd",
+          "flake-compat"
+        ],
+        "nix2container": [
+          "nixbsd"
+        ],
+        "nixpkgs": [
+          "nixbsd",
+          "nixpkgs"
+        ],
+        "nixpkgs-regression": [
+          "nixbsd"
+        ],
+        "pre-commit-hooks": [
+          "nixbsd"
+        ]
+      },
+      "locked": {
+        "lastModified": 1741969094,
+        "narHash": "sha256-N9URV6eRf3FfWPuFHwEpTjrDBwvP9ekNPeMe5HqPycY=",
+        "ref": "freebsd-build",
+        "rev": "d003841587befb47d41373cba00d9921df8e237b",
+        "revCount": 17650,
+        "type": "git",
+        "url": "https://git.lix.systems/artemist/lix.git"
+      },
+      "original": {
+        "ref": "freebsd-build",
+        "type": "git",
+        "url": "https://git.lix.systems/artemist/lix.git"
+      }
+    },
+    "mini-tmpfiles": {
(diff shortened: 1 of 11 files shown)


rhelmot:
`Invalid fstype: Invalid argument`

I think you need `kldload /run/current-system/kernel-modules/nullfs.ko`. That should probably be configured automatically...

zowoq:
> I think you need `kldload /run/current-system/kernel-modules/nullfs.ko`.

That works, thanks!

rhelmot:
It looks like the stdenv fixes are going to land soon, anything I can do to help move this along?

rhelmot:
What are the reproduction instructions? I am currently building a lot of things, both in a vm and on actual hardware.

If it's complicated, can you just tell me which binary is segfaulting?

rhelmot:
The most likely culprit is probably that you're trying to run a linux binary on FreeBSD somehow

zowoq:
Seems building inside the vm works but get a segfault when using the vm as a remote builder. Just trying to build `hello` or anything else via remote seems to trigger it.

https://github.com/nixos-bsd/nixbsd/tree/tmp3

Is the tmp3 branch useable? The new `virtualisation.rootSize` option which would be useful, could drop the nullfs fix I have here as well.

zowoq:
> Changing `--system` to `--eval-system` makes everything work perfectly.

Yes, that works, thanks. 

I've switched to the tmp3 branch and updated the nixpkgs input.

For now I'll deploy just the vm on community build box and later when the stdenv fixes in nixpkgs land I'll deploy the CI vm.

For access to the community build box and vm add your user and key in this file:
https://github.com/nix-community/infra/blob/master/modules/nixos/community-builder/users.nix

zowoq:
At this time I don't want to make any sort of announcement to the nix community about freebsd availability, for now I'd say that this is just for the nixbsd devs to use.

cc @zimbatm

PR #1759, opened 2025-04-03
switch to dnscontrol

zowoq:
Previously looked at stateless dns with octodns as part of NixOS-DNS in https://github.com/nix-community/infra/pull/1297, dnscontrol seems to be a bit nicer to use even with the js dsl.

I think the dns records are the important part of our terraform state, by managing them with another (stateless) tool it makes experimenting and moving off terraform + terraform cloud simpler.

Thought about adding dnscontrol preview/push github actions but I'd rather just manage it manually for now.


dnscontrol/creds.json
```diff
@@ -0,0 +1,7 @@
+{
+  "cloudflare": {
+    "TYPE": "CLOUDFLAREAPI",
+    "accountid": "e4a2db52c495db230973c839a0699ae1",
+    "apitoken": "$CLOUDFLARE_API_TOKEN"
```
zowoq, line 5: I thought we could use our own tokens instead of keeping the super admin account token in sops but don't feel strongly about it.

dnscontrol/dnsconfig.js
```diff
@@ -0,0 +1,106 @@
...
+    DnsProvider(DSP_CLOUDFLARE),
+
+    records,
+
+    // blocks other CAs from issuing certificates for the domain
+    CAA("@", "issue", "letsencrypt.org"),
+
+    // "cannot create CNAME record for bare domain", ALIAS -> CNAME
+    ALIAS("@", "nix-community.github.io."),
+    CNAME("www", "nix-community.github.io."),
```
zowoq, line 90: This www cname is an extra record that wasn't in our terraform config.

PR #1763, opened 2025-04-05
terraform: switch to opentofu and postgresql backend, drop terraform cloud

zowoq:


terraform/terraform_backend.tf
```diff
@@ -1,6 +1,18 @@
 terraform {
-  backend "remote" {
-    organization = "nix-community"
-    workspaces { name = "nix-community" }
+  # will import resources into new project, won't transfer existing state
+  backend "s3" {
+    endpoint = "s3.nix-community.org"
```
Mic92, line 4: Don't know how much state churn we deal with, but I am also now quite happily using the encrypted state backend that can be committed to git. Works fine for small projects.

zowoq: Removing dns means there isn't much churn apart from occasional changes to hydra so committing state would probably be fine for this repo but I'm hoping we can also use it for https://github.com/ipdxco/github-as-code.

IIRC opentofu is still going to implement OCI for state storage so s3 may only be an interim solution anyway.

Mic92:
Fine with me

zimbatm:
is the terraform state backed up? my only concern is that if web02 disappears, it will be some work to recover to re-create web02 and re-import the state.

zowoq:
I've removed all of the github and hydra resources from the previous terraform cloud state but I've left the tfe resources intact so if we need to switch back to terraform cloud we don't need to set it from scratch.

PR #1776, opened 2025-04-12
build01: switch to new hardware

zowoq:
Fundraising target is at https://opencollective.com/nix-community.


zowoq:
With the hetzner price increases we're probably not going to upgrading this machine.

PR #1794, opened 2025-04-22
use ssh for sops keys

zowoq:
~~Can't do the same for admins yet as currently I'm using `SOPS_AGE_KEY` and there isn't an equivalent for ssh, it only supports paths.~~

Can have both and have a check to ensure it is up to date.

### DIFF
--- a/dev/shell.nix
+++ b/dev/shell.nix
@@ -9,7 +9,6 @@
           python3.pkgs.deploykit
           python3.pkgs.invoke
           sops
-          ssh-to-age
           yq-go
         ];
       };
--- a/devdoc/onboarding.md
+++ b/devdoc/onboarding.md
@@ -2,9 +2,9 @@
 
 - Add them to the [list of administrators](../docs/administrators.md)
 
-- Add their user and ssh key to [users](../users) as member of the `trusted` and `wheel` groups.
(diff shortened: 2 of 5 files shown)


dev/sops.nix
```diff
@@ -0,0 +1,55 @@
...
+            EOF
+            git init --quiet
+            git add .
+            git commit -m init --quiet
+            export LANG=${if pkgs.stdenv.isDarwin then "en_US.UTF-8" else "C.UTF-8"}
+            export LC_ALL=${if pkgs.stdenv.isDarwin then "en_US.UTF-8" else "C.UTF-8"}
+            export NIX_STATE_DIR=$TMPDIR/state NIX_STORE_DIR=$TMPDIR/store
+            inv update-sops-files || true
+            git status --short
+            git --no-pager diff --exit-code
```
Mic92, line 51: This shouldn't be able to actually change any key because `inv update-sops-files` is missing a sops key to actually update the encryption.

zowoq: Yes, update-sops-files won't work. Updated to separate the steps, it'll error if .sops.yaml has a diff and it'll error if any of any of the yaml secrets need to be updated.

PR #1796, opened 2025-04-24
modules/shared: fix host-info

zowoq:

### DIFF
--- a/modules/shared/host-info.bash
+++ b/modules/shared/host-info.bash
@@ -1,4 +1,4 @@
-flake=$(nix flake metadata self --json | jq -r '.path' | sed -e 's|/nix/store/||' -e 's|-source||')
+flake=$(nix registry list | grep 'flake:self' | cut -d ':' -f3 | sed -e 's|/nix/store/||' -e 's|-source||')
 nix_version="$(nix store ping --store daemon --json | jq -r '.version')"
 case "$(uname -s)" in
 Darwin)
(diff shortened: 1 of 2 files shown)


modules/shared/host-info.bash
```diff
@@ -1,4 +1,4 @@
-flake=$(nix flake metadata self --json | jq -r '.path' | sed -e 's|/nix/store/||' -e 's|-source||')
```
zowoq, line 1: @Mic92 this was broken by the 2.28.2 update.
```
$ nix flake metadata self
error: cannot find flake 'flake:self' in the flake registries

$ nix registry list
system flake:nixpkgs path:/nix/store/pp22j0fx88a4dnz06z0hli0kaj0jsv4n-source
system flake:self path:/nix/store/0jd6nr3xn1ib9jb886723isclvaz04cy-source
```

Mic92: @zowoq did 2.28.1 still works?

Mic92: Haven't yet tested, but I suspect this is fixed by? https://github.com/NixOS/nix/pull/13076

Mic92: Seems to fix the issue.

zowoq: Yes, that fixes it, thanks. Sorry for missing the issue that was opened a couple of days ago.

PR #1807, opened 2025-05-01
build06: init

zowoq:
Will wait until the config is finalised before ordering the machine. 

cc @ConnorBaker @SomeoneSerge 

This wires the machine up so it can be used by https://hydra.nix-community.org/ (and https://buildbot.nix-community.org/) but only for builds that set the `cuda` requiredSystemFeature.


hosts/build06/nvidia.nix
```diff
@@ -0,0 +1,18 @@
+{
+  hardware.graphics = {
+    enable = true;
+  };
+
+  hardware.nvidia = {
+    open = true;
+  };
+
+  programs.nix-required-mounts = {
```
SomeoneSerge, line 10: We'll need to hack around https://github.com/NixOS/nix/issues/9272, my tentative plan was to see if it's enough to ad hoc patch Nix just on the remote side without touching the requesting side

zowoq: We can try that but as long as it can eventually be upstreamed patching both sides is fine.

@Mic92 What do you think a fix for https://github.com/NixOS/nix/issues/9272 would look like?

zowoq: As the only builds that'll run on this machine are the cuda tests could we set the sandbox paths directly instead of using the hook?

Mic92: What is the issue here? Is this needed for some cachix pre-build hook?

zowoq: > What is the issue here? Is this needed for some cachix pre-build hook?

Did you read the issue linked in the previous comments?

SomeoneSerge: A simpler alternative is to relax the sandbox unconditionally

zowoq: Can you clarify what unconditionally means? Adding the paths that would have been set by the pre-build-hook to `nix.settings.sandbox-paths` or just setting `nix.settings.sandbox = false`?

modules/nixos/hydra.nix
```diff
@@ -51,6 +51,7 @@ in
         printf "$machines" > $out
         substituteInPlace $out --replace-fail 'ssh-ng://' 'ssh://'
         substituteInPlace $out --replace-fail ' 80 ' ' 3 '
+        substituteInPlace $out --replace-fail ' 14 ' ' 2 '
```
ConnorBaker, line 54: This should change depending on https://github.com/nix-community/infra/pull/1807/files#r2075684863, correct?

hosts/build06/nvidia.nix
SomeoneSerge: > Adding the paths that would have been set by the pre-build-hook to nix.settings.sandbox-paths

This. My concern with that is, can't it happen that you have a `requiredSystemFeatures = [ "cuda" ]` derivation scheduled to `build06` that depends on an unbuilt ...`torch`, which is accidentally referenced from the outputs, and then `build06` ends up building a non-test derivation with a relaxed sandbox and sending the result back to the hydra machine and ultimately leaking the less reproducible non-test output to cachix?

I'm a bit hazy on the hydra+distribuitedBuilds architecture/interactions, wdyt? We could also say that this is a non-concern because it'll get GC-ed

hosts/build06/default.nix
```diff
@@ -0,0 +1,18 @@
...
+    # no hercules CI, doesn't support mandatoryFeatures
+    ./nvidia.nix
+    inputs.self.nixosModules.cgroups
+    inputs.self.nixosModules.ci-builder
+    inputs.self.nixosModules.disko-zfs
+    inputs.self.nixosModules.remote-builder
+    inputs.srvos.nixosModules.hardware-hetzner-online-intel
+  ];
+
+  nix.settings.max-jobs = 14;
```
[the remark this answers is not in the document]
SomeoneSerge: 👍🏻 this is the place where we suffer very much from lack of a "real" scheduler for Nix, with support for negative affinity and resource constraints like in SLURM. We'll definitely run into issues running things like pytorch or pytorch-lightning test-suites, where you can have matrices of tests run in parallel

hosts/build06/nvidia.nix
zowoq: AFAIK your concern is accurate. 

I think that means we're blocked from moving forward with this until we have some sort of solution for https://github.com/NixOS/nix/issues/9272.

zowoq: Could we force the build and test derivations to run on different hosts if we also set `requiredSystemFeatures` for the build derivation and remove that feature from build06?

| derivation | big-parallel | cuda |
|:-:|:-:|:-:|
| build |  ✅ ||
| test || ✅ |

| host | big-parallel | cuda |
|:-:|:-:|:-:|
| build03 (epyc) |  ✅ ||
| build06 (gpu) || ✅ |

SomeoneSerge: > I think that means we're blocked from moving forward with this

OTOH do we actually care? These outputs get garbage collected anyway.

> Could we force the build and test derivations to run on different hosts if we also set requiredSystemFeatures for the build derivation and remove that feature from build06?

Adding features to _non-test_ derivations instead of the test derivations is an interesting idea. However, a derivation's system features are part of the derivation struct and affect its hash

zowoq: > OTOH do we actually care? These outputs get garbage collected anyway.

Everything built gets pushed to our cachix cache. To avoid that we'd need to have a separate hydra instance just for running the tests. That then has problems, either we'd need to build the non-test derivations twice or have the test derivations on a much slower schedule to ensure that they have already been build on the main hydra (and still have the problem of derivations that failed on the main hydra being attempted a second time on the test hydra).

> Adding features to _non-test_ derivations instead of the test derivations is an interesting idea. However, a derivation's system features are part of the derivation struct and affect its hash

Yes. As long as the feature is a nix default (which big-parallel is) and also correct (I think big-parallel does apply for most of them) I don't see an issue?

SomeoneSerge: > Everything built gets pushed to our cachix cache.

What I meant to say is: cachix contents eventually get garbage-collected...

SomeoneSerge: > a nix default (which big-parallel is) and also correct (I think big-parallel does apply for most of them) I don't see an issue?

Ah, I see. Clever. This should work?

zowoq: https://github.com/helsinki-systems/hydra-queue-runner, https://github.com/NixOS/hydra/pull/1497/files

Using the new queue runner seems like it would address this problem.

zowoq:
How are these tests going to be set up in nixpkgs for CI to run? Will they be added to the current release-cuda jobset or a new jobset?

zowoq:
> > How are these tests going to be set up in nixpkgs for CI to run? Will they be added to the current release-cuda jobset or a new jobset?
> 
> @ConnorBaker @SomeoneSerge

^ asking again

zowoq:
https://github.com/nix-community/infra/pull/1912 is merged but I want to see that it can handle a full rebuild from a staging-next merge without problems. There is a staging-next PR open at the moment which I expect will be merged in a week or two, after that should be ready to move forward with this.

zowoq:
> #1912 is merged but I want to see that it can handle a full rebuild from a staging-next merge without problems. There is a staging-next PR open at the moment which I expect will be merged in a week or two, after that should be ready to move forward with this.

Staging-next was merged earlier, there wasn't any problems with the new queue runner handling the full rebuild so we're ready to move forward with this. 

I'll wait for the tests to be included in a jobset that we can run before ordering the machine.

zowoq:
https://discourse.nixos.org/t/nix-flox-nvidia-opening-up-cuda-redistribution-on-nix/69189

> https://matrix.to/#/!eWOErHSaiddIbsUNsJ:nixos.org/$jZY-rPxCzGokgIGyZGMHXdP-iM2-LLQTMoycDZ6_J5g?via=nixos.org&via=matrix.org&via=nixos.dev
> The negotiations with NVIDIA have been run by Flox (although in parallel with many other companies' simultaneous inquiries). Ron kept us, the Foundation, and the SC in the loop, and offered both legal help and workforce. The current idea roughly is that the CUDA team gets access to the relevant repo and infra, and work closely together with Flox to secure the position and a commx channel to NVIDIA. 

I've been seeing occasional issues from people about cache misses because we're hitting the limits of our cache capacity, we really need to significantly increase our cache capacity (https://github.com/nix-community/infra/issues/1926) for the cuda builds and tests to be sustainable.

Not going to waste community money on a cache we don't need if flox is already doing it.

I don't think we should proceed with the gpu tests here and we should also stop building the cuda packages.

SomeoneSerge:
> Honestly can't believe that no one bothered to mention that flox was working on their own cache and infra before the announcement.

DISCLAIMER: I didn't check the latest revision of this message with Connor or anyone else, and so cannot represent anyone but myself.


Hello @zowoq! This is a failure, on part of the cuda team at least. I can see how this breaks trust or reads as some sort of disrespect. I hope we can one day sit in one room none the less.
Regarding the jobset, if you'd rather separate and make distance, we can start running our own. We wouldn't want to abuse your work. We still do aim to implement GPU Tests outside Flox, for one thing because we're not currently aware of Flox having any interest in GPU Tests, but mainly because a Nix-Community-like structure offers a clearer story for sponsorships, greater flexibility, and more diversification. Concerning Cachix, I believe that to split and host that under a new name would involve too much false-signaling and communication overhead. It would be just more pragmatic if we could reuse the Nix-Community substitutor and include its scaling in the GPU Tests' "budget". Of course a dedicated server or a backblaze bucket used as a backend for snix-narbridge or attic must be still an order of magnitude more affordable than Cachix, as you outline in https://github.com/nix-community/infra/issues/1926.

With that out of the way, I'd like to attach a brief transcript of interactions between the CUDA Team and Flox to date so you may the better judge whether to cooperate with us any further.

0. On May 5 Foundation board member Sebastian (ra33it) and the CUDA Team started a private matrix room for coordinating the communications that various parties (including commercial entities) have been engaging in, in regards to the CUDA EULA issue. We only included people we were already talking to privately and had consented to participating in a group chat. We also invited invited Ron and Tom as representatives of the Board and the SC respectively, and of Flox both. Ron shared that they at Flox had been very interested in the issue "for years", and are trying to find relevant contacts.
1. On June 13 Ron, Connor, and I had a group call.
   - Connor and I provided a recap on the team's status with the infrastructure, labour, and communication with the potential sponsor companies.
   - Ron shared that they at Flox had managed to meet a contact at NVIDIA, which could potentially lead to progress for the EULA issue. A Flox-hosted cache was mentioned.
   - We all expressed an agreement that in it's in everyone's long-term interst to see Nix CUDA cache outside Flox, preferably under the Foundation.
   - We also discussed the concept of a federated community infrastructure and of sponsoring third party non-profits (like hackspaces) to maintain their own and depend on the Foundation less. I believe Jonas was directly referenced.
   - We all acknowledged that, after years of NVIDIA not responding to any attempts at starting a conversation on part of various Nixpkgs contributors and their companies, securing any communication channel at all is a priority.
2. On June 16 Ron shared an address he would present to the SC. The letter addressed the hypothetical outcome of NVIDIA only allowing Flox (a US-based for-profit entity) to redistribute "patchelfed" CUDA. None of this was to be publicly disclosed, as no deal had been secured by then, and neither company made any public announcements.
3. By June 24th the CUDA team sent back its feedback for the letter, which included minor stylistic edits, but also general comments about conflicts of interest, sustainability, cost of labour, interactions between the Foundation, the SC, and "the industry", etc.
4. On July 8 Ron shared that they are beginning to "draft contracts". I'm going to guess that by then the SLA-based arrangement was basically confirmed, and we must have known that Flox would have to do the redistribution via their own cache.
5. Between August 14 and August 19, Ron and Sebastian suggested that an announcement slide be added to the State of the Union presentation at NixCon.
6. On August 24 I, Connor, and Ron had another call about who and what is it acceptable to mention during State of the Union. Ron also shared that they might not be able to make the announcement at NixCon because of NVIDIA postponing theirs.
7. On September 5 Ron confirmed that they cannot make the announcement at NixCon.
8. On September 9 Ron shared they were given the green light and have to act quick to make the announcement.

Hope this gives you some answers. In particular, yes, we had known that the potential Flox-NVIDIA arrangement would involve a binary cache, and it was an oversight not to include you in the discussion back then. While I'm quite happy that Flox is open to talking to and cooperating with the Nixpkgs developer community, Flox's binary cache is, at the end of the day, Flox's binary cache. Flox's private deals with NVIDIA are Flox's private deals with NVIDIA. I personally do not think the Flox-NVIDIA partnership is any reason at all to abandon the CUDA CI efforts under Nix-Community. The communication and trust issues, of course, would be. If you do find yourself still willing to talk to us, both we at the CUDA Team and, I was just assured, Ron on the Flox side would appreciate having a meeting with you.

Cheers,
Serge

zowoq:
Rather than responding to all that immediately I'd like an answer to another question first. If you scroll up to https://github.com/nix-community/infra/pull/1807#issuecomment-3141631139 you'll see that I asked the same question three times over the course of a month before I got a response, why wasn't it answered promptly?

PR #1809, opened 2025-05-04
modules/nixos/cgroups: patch nix

zowoq:

### DIFF
--- a/modules/nixos/cgroups.nix
+++ b/modules/nixos/cgroups.nix
@@ -1,5 +1,16 @@
+{ pkgs, ... }:
 {
   nix = {
+    package = pkgs.nix.overrideAttrs (o: {
+      patches = (o.patches or [ ]) ++ [
+        (pkgs.fetchpatch {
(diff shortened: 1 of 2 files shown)


modules/nixos/cgroups.nix
```diff
@@ -1,5 +1,16 @@
+{ pkgs, ... }:
 {
   nix = {
+    package = pkgs.nix.overrideAttrs (o: {
+      patches = (o.patches or [ ]) ++ [
+        (pkgs.fetchpatch {
+          # https://github.com/NixOS/nix/pull/13135
+          url = "https://github.com/NixOS/nix/commit/7030d2e44fccc6c0fc5f5bc87eb05d7134744131.patch";
```
Mic92, line 8: I think we might have to revert this because I saw that the waitpid() will steel the process status in other places.

zowoq: https://github.com/nix-community/infra/pull/1818

Issue #1007, opened 2023-12-05
native `ssh` daemon on darwin hosts lacks fido support


zowoq:
This blocks community members from using yubikeys, etc.

Mic92:
Does nix-darwin offers a replacement? What does actually macOS upgrades look like, do they need the kvm console?

zowoq:
> Does nix-darwin offers a replacement?

No. 

> What does actually macOS upgrades look like

Usually simple but if it doesn't work we'll likely be locked out.

> do they need the kvm console?

Don't know if hetzner actually anything like this for macos, they only mention ssh or vnc from the system.

https://docs.hetzner.com/robot/dedicated-server/mac-mini/getting-started/

Mic92:
I think a colleagues of mine did try to use kvm.

zowoq:
Looks like robot.hetzner.com will let me request remote console for the macs.

zowoq:
I haven't really got around to investigating this yet anyway, I don't know yet if just upgrading the system would be enough to resolve this or if it just gets us a newer ssh version and we'd still need to provide the fido libs or whatever.

Mic92:
@phear does macos has now ssh fido support?

zowoq:
```
hetzner@darwin03> sw_vers                                                                                                                                                                                
ProductName:            macOS
ProductVersion:         14.3
BuildVersion:           23D56
```

Needed a kvm console to restore vnc on darwin03 so I upgraded the macos version as well. If we still need to provide the fido libs I'll see if I can sort it out in the next couple of weeks.

zowoq:
Tentatively fixed in https://github.com/nix-community/infra/pull/1527.

thunze:
FWIW, I was able to log in with a `sk-ecdsa-sha2-nistp256@openssh.com` key today. See https://github.com/nix-community/infra/pull/2403.

zowoq:
Thanks for reporting that it works.

PR #1824, opened 2025-05-11
users/keys/mic92: switch to hardware based keys

Mic92:


users/keys/mic92
```diff
@@ -1 +1,2 @@
-ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIKbBp2dH2X3dcU1zh+xW3ZsdYROKpJd3n13ssOP092qE joerg@turingmachine
+sk-ssh-ed25519@openssh.com AAAAGnNrLXNzaC1lZDI1NTE5QG9wZW5zc2guY29tAAAAIEVSsc5mlP8aWiUVwWWM3gKlB5LHVpmKSifnDyox/BnVAAAABHNzaDo= yubikey1
```
zowoq, line 1: https://github.com/nix-community/infra/blob/544c402c0a8de54820f7f2181a8d35610f6933f7/modules/darwin/community-builder/default.nix#L15

Security keys don't work ootb on darwin so I added support just for the darwin build box but we haven't confirmed that it works yet. If it does work we can move it to the common config so your key will work on both darwin hosts.

zowoq (commented):
Does your age key need to be updated as well?

users/keys/mic92
Mic92: Ok. I can test. Otherwise I can also switch to my tpm ssh key.

Mic92:
> Does your age key need to be updated as well?

Later. Haven't migrated it yet.

users/keys/mic92
Mic92: Key actually doesn't work. Will switch to my tpm key instead.

zowoq: Thanks for testing. I've reverted the SecurityKeyProvider and reopened the original issue.

https://github.com/nix-community/infra/issues/1007, https://github.com/nix-community/infra/pull/1825

PR #1837, opened 2025-05-20
build03: add rfc39

zowoq:

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -44,6 +44,8 @@
     nixpkgs.url = "git+https://github.com/NixOS/nixpkgs?shallow=1&ref=nixos-unstable-small";
     nur-update.inputs.nixpkgs.follows = "nixpkgs";
     nur-update.url = "github:nix-community/nur-update";
+    rfc39.inputs.nixpkgs.follows = "nixpkgs";
+    rfc39.url = "github:NixOS/rfc39";
     sops-nix.inputs.nixpkgs.follows = "nixpkgs";
     sops-nix.url = "github:Mic92/sops-nix";
     srvos.inputs.nixpkgs.follows = "nixpkgs";
--- a/modules/nixos/rfc39.bash
+++ b/modules/nixos/rfc39.bash
@@ -1,0 +1,50 @@
+set -eux
+
+export GIT_SSH_COMMAND="ssh -i $RFC39_RECORD_SSH_KEY"
+export GIT_AUTHOR_NAME="rfc39"
+export GIT_AUTHOR_EMAIL="rfc39@nix-community"
+export GIT_COMMITTER_NAME="rfc39"
+export GIT_COMMITTER_EMAIL="rfc39@nix-community"
+
+recordsdir=$HOME/rfc39-record
+if ! [[ -e $recordsdir ]]; then
+  git clone git@github.com:nix-community/rfc39-record.git "$recordsdir"
+fi
+cd "$recordsdir"
+git fetch origin --no-auto-maintenance
+git checkout main
+git reset --hard origin/main
+git maintenance run --auto
+
+repos=(
+  "home-manager all-maintainers.nix 13467276"
+)
+
+for r in "${repos[@]}"; do
+  read -r REPO MAINTAINER_LIST TEAM <<<"$r"
+
+  DIR=$HOME/$REPO
+
+  if ! [[ -e $DIR ]]; then
+    git clone "https://github.com/nix-community/$REPO.git" "$DIR"
+  fi
+  cd "$DIR"
+  git fetch origin --no-auto-maintenance
+  git checkout origin/HEAD
+  git maintenance run --auto
+
+  rfc39 \
+    --credentials "$RFC39_CREDENTIALS" \
+    --maintainers "$MAINTAINER_LIST" \
+    sync-team nix-community "$TEAM" \
+    --limit 10 \
+    --invited-list "$recordsdir/$REPO-invitations"
+done
+
+cd "$recordsdir"
+
+if [ -n "$(git status --porcelain)" ]; then
+  git add .
+  git commit -m "Automated team sync results."
+  git push origin main
+fi
--- a/.sops.yaml
+++ b/.sops.yaml
@@ -106,6 +106,15 @@ creation_rules:
     path_regex: ^modules/secrets/hercules-ci.yaml$
   - key_groups:
       - age:
+          - age1qg7tfjwzp6dxwkw9vej6knkhdvqre3fu7ryzsdk5ggvtdx854ycqevlwnq
+          - age1dzvjjum2p240qtdt2qcxpm7pl2s5w36mh4fs3q9dhhq0uezvdqaq9vrgfy
+          - age17n64ahe3wesh8l8lj0zylf4nljdmqn28hvqns2g7hgm9mdkhlsvsjuvkxz
+          - age1d87z3zqlv6ullnzyng8l722xzxwqr677csacf3zf3l28dau7avfs6pc7ay
+          - age1jrh8yyq3swjru09s75s4mspu0mphh7h6z54z946raa9wx3pcdegq0x8t4h
+          - age1m7xhem3qll35d539f364pm6txexvnp6k0tk34d8jxu4ry3pptv7smm0k5n
+    path_regex: ^modules/secrets/rfc39_private_key.der$
+  - key_groups:
+      - age:
           - age1dzvjjum2p240qtdt2qcxpm7pl2s5w36mh4fs3q9dhhq0uezvdqaq9vrgfy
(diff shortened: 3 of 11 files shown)


zowoq:
Okay, this seems to work, I tried a dry run of the bot on the stylix repo and it attempted to add people to `@nix-community/stylix-maintainers`.

@danth @khaneliman The bot reports errors via prom metrics but the way it is implemented is rather awkward and frankly isn't worth the hassle (https://github.com/NixOS/rfc39/issues/14). Even if I do wire it to our monitoring and route a notification to you all it will tell you is that it started failing, nothing more specific. The logs have more detail but I can't make them public as it can leak the bots secrets.

https://github.com/NixOS/nixpkgs/blob/master/lib/tests/maintainers.nix

I think you're going to need to run the nixpkgs maintainers validation checks as a PR check in your repos to catch errors.

flake.nix
```diff
@@ -36,6 +36,8 @@
     nixpkgs.url = "git+https://github.com/NixOS/nixpkgs?shallow=1&ref=nixos-unstable-small";
     nur-update.inputs.nixpkgs.follows = "nixpkgs";
     nur-update.url = "github:nix-community/nur-update";
+    rfc39.inputs.nixpkgs.follows = "nixpkgs";
+    rfc39.url = "github:qowoz/rfc39/flake-utils";
```
Mic92, line 40: Can merge fixes upstream in rfc39 btw.

zowoq:
Seems incorrect quoting will break it completely.

```
called `Result::unwrap()` on an `Err` value: Serde(Error("invalid type: string \"179992797\", expected u64", line: 1, column: 81))
```
```diff
diff --git a/modules/lib/maintainers.nix b/modules/lib/maintainers.nix
index 4684ad38..69e428cb 100644
--- a/modules/lib/maintainers.nix
+++ b/modules/lib/maintainers.nix
@@ -11,7 +11,7 @@
     name = "Eulalia del Sol";
     email = "3ulalia@proton.me";
     github = "3ulalia";
-    githubId = "179992797";
+    githubId = 179992797;
   };
   "9p4" = {
     name = "9p4";
@@ -268,7 +268,7 @@
   henrisota = {
     email = "henrisota@users.noreply.github.com";
     github = "henrisota";
-    githubId = "56848082";
+    githubId = 56848082;
     name = "Henri Sota";
   };
   hey2022 = {
@@ -558,7 +558,7 @@
   msfjarvis = {
     email = "me@msfjarvis.dev";
     github = "msfjarvis";
-    githubId = "13348378";
+    githubId = 13348378;
     name = "Harsh Shandilya";
     keys = [
       {

```

Maintainers missing information are skipped.

```
WARN Missing GitHub ID, github_account: hey2022, exec-mode: CheckHandles, module: rfc39::op_check_handles:17
ERRO Missing GitHub Account, but ID present, github_id: 43591752, who: jess, exec-mode: CheckHandles, module: rfc39::op_check_handles:21
WARN Missing GitHub ID, github_account: silmarp, exec-mode: CheckHandles, module: rfc39::op_check_handles:17
```

zowoq:
I didn't notice that the maintainer lists are deduplicated with nixpkgs maintainers, rfc39 only works for maintainers that are listed in the file. Would need to add an entry in the maintainer file, e.g. `inherit (pkgs.lib.maintainers) zowoq;` for nixpkgs maintainers. @danth @khaneliman Is that feasible or do we need to look for a different method of doing the invites?

zowoq:
```
rfc39 \
    --maintainers ${./maintainers.nix} \
    --maintainers ${nixpkgs}/maintainers/maintainer-list.nix
```

Using the nixpkgs maintainer list will invite everyone from nixpkgs.

```
rfc39 \
    --maintainers ${./maintainers.nix} \
    --maintainers ${./maintainers-nixpkgs.nix}
```

If you meant something like this, I don't see how that is any different from a single file? There will still be the boilerplate and both files will need a validation check, if a someone is removed from the upstream maintainer list in nixpkgs it'll break the downstream list and the rfc39 invites.

MattSturgeon:
Ah I see. I was thinking of it kinda backwards.

In that case I think what we want is for projects that don't want to explicitly list all their maintainers to _generate_ a list of their maintainers.

E.g.

```nix
let
  # Assume this configuration has a `meta.maintainer` option that maps filename→maintainer
  configuration = /* some module eval */;

  # Collect all the module maintainers
  maintainers = lib.pipe configuration.config.meta.maintainer [
    builtins.attrValues
    lib.unique
    (builtins.filter (m: m ? github))
    (map (m: lib.nameValuePair m.github m))
    builtins.listToAttrs
  ];
in

# return a JSON file
writers.writeJSON "maintainers.json" maintainers
```

I.e. it is up to the downstream project to provide a list of maintainers to invite, either using an in-repo file (as-per nixpkgs) or a generated file (as above)

zowoq:
> either using an in-repo file (as-per nixpkgs) or a generated file (as above)

I think generated file would need to be nix but yes, either would work. The downstream project would still need to ensure that the generated file passes the validation check.

khaneliman:
Basically, we just need an exported file that contains all the maintainers that would possibly be pinged so that they can be invited to the nix-community team for unprivileged maintainers? That should be doable.

modules/nixos/rfc39.bash
```diff
@@ -0,0 +1,51 @@
...
+cd "$recordsdir"
+git fetch origin --no-auto-maintenance
+git checkout main
+git reset --hard origin/main
+git maintenance run --auto
+
+repos=(
+  #"home-manager modules/lib/maintainers.nix ?"
+  "stylix stylix/maintainers.nix 13054517"
+)
```
MattSturgeon, line 22: Could we make it so that a repo entry can some arbitrary command to run? Or some file to `nix-instantiate --eval`? Or some flakeref to `nix eval`?

Assuming you want an _eval_ that spits out a nix value, rather than a _build_ that builds a file.

E.g. nixvim might want something like `nix-instantiate --eval ./.github/list-maimtainers.nix` which would internally use flake-compat to load nixvim and would then scan an empty `nixvimConfiguration` for plugin maintainers.

I assume stylix would do something similar, scanning their `modules/«module»/meta.nix` files.

IIRC home-manager uses a similar `meta.maintainers` module to NixOS, so they would probably get their maintainers from an empty `homeConfiguration`.

zowoq: Downstream projects will already need to eval and validate the file to ensure that it isn't going to break, just check the file into the repo, don't make it more complicated then it needs to be by doing evals here as well. If multiple projects are going to be using this bot I want this side to be kept as simple as possible.

MattSturgeon: If the file has to be checked into the repo anyway, then I don't see any value in generating it dynamically.

---

However, does rfc39 even support evaluating a list that is actually a function?

I.e. this is different from what it is used to consuming from nixpkgs:

```nix
{ lib }:
{
  # ...
}
```

---

It would also be important to ensure the correct nixpkgs revision is used, in case a maintainer was recently added or removed.

In that case should downstream maintainer lists be reading their lockfile to download the correct nixpkgs?

Does rfc39 support a "real" nix eval of the maintainers list file? Would rfc39 support evaluating a "compatibility wrapper" that imports the real list internally?

```nix
# maintainers-list-compat.nix
let
  lockFile = builtins.fromJSON (builtins.readFile ../flake.lock);
  nixpkgs = builtins.fetchTarball {
    # ...
  };
  lib = import (nixpkgs + "/lib");
in
import ./maintainer-list.nix { inherit lib; }
```

zowoq: Again, this is making it more complicated then it needs to be and pushing potential problems upstream, I want this side to be as simple as possible.

MattSturgeon: > Again, this is making it more complicated then it needs to be and pushing potential problems upstream, I want the infra side to be as simple as possible.

Maybe I wasn't clear - this _would_ be downstream code.

I'm only asking if the rfc39 tool evaluates the maintainers-list file using nix (or lix, etc) or if it uses some other static analysis tool like rnix. I.e. whether it'd support evaluating a _downstream_ wrapper expression that fetches the right nixpkgs and passes it to a maintainers list file internally.

If the file can't be evaluated "properly", then downstream projects can't inherit maintainers from nixpkgs `lib.maintainers` as [suggested earlier](https://github.com/nix-community/infra/pull/1837#issuecomment-2897445559).

zowoq: Sorry, seems I wasn't clear.

No wrappers, generate the file downstream and check it in to the repo.

> If the file can't be evaluated "properly", then downstream projects can't inherit maintainers from nixpkgs `lib.maintainers` as [suggested earlier](https://github.com/nix-community/infra/pull/1837#issuecomment-2897445559).

Disregard this suggestion.

06kellyjac:
I can report it worked as I was invited to `@nix-community/home-manager-maintainers` this morning to my surprise :slightly_smiling_face:

PR #1842, opened 2025-05-26
modules/nixos/community-builder/users: allow wheel group

zowoq:
Currently the nix-community admins share the admin account on the darwin boxes, for now I'll just manage the sudo permissions manually for other users.

### DIFF
--- a/modules/darwin/community-builder/users.nix
+++ b/modules/darwin/community-builder/users.nix
@@ -256,6 +256,7 @@ let
       # lib.maintainers.emily, https://github.com/emilazy
       name = "emily";
       trusted = true;
+      # wheel = true; admin group set manually
       uid = 542;
       keys = ./keys/emily;
     }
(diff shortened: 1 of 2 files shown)


zowoq:
@emilazy I'm still looking at the best way of handling the remote builds max-jobs issue but I thought this might help, with sudo you'd be able post to wall and to kill builds that are overloading the host.

zowoq:
merged cores / max-jobs in https://github.com/nix-community/infra/commit/29564d96c8811f8c8c6599a7f2b9af0b94c2fb8d

emilazy:
Sorry for missing this! Ideally this wouldn’t be necessary but if you want to entrust me to send `wall` messages / kill runaway build jobs then I’m happy to commit to using the powers for that and nothing else.

zowoq:
A couple of times I've found that they can be almost non-responsive making it difficult to identify and kill specific jobs, feel free to just do a reboot if that is easier.

PR #1844, opened 2025-05-31
build02: disable auto-optimise-store

zowoq:
I think auto-optimise-store is too efficient on this machine, gc is very slow (days compared to hours on the others), try running without it a few weeks.

### DIFF
--- a/hosts/build02/default.nix
+++ b/hosts/build02/default.nix
@@ -21,9 +21,6 @@
   nix.settings.cores = config.nix.settings.max-jobs / 3 * 2;
   nix.settings.max-jobs = 24;
 
-  # set in srvos, remove when reinstalling
-  networking.hostId = "deadbeef";
-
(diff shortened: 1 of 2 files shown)


hosts/build02/default.nix
```diff
@@ -11,6 +11,8 @@
     inputs.srvos.nixosModules.hardware-hetzner-online-amd
   ];
 
+  nix.settings.auto-optimise-store = lib.mkForce false;
```
Mic92, line 14: My experience has been that is keep-derivations that slows down gc quite a bit:

sudo nix-collect-garbage -d --option keep-derivations false

zowoq: I'll let it fill up and try that.

zowoq: `keep-derivations false` helped a little but still very slow, I forgot to copy the exact gc output but hardlinking was saving a lot of space. I thought I'd reinstall as it would be quicker than gc. nixos-anywhere is now very fast, only took about five minutes.

PR #1846, opened 2025-05-31
nixos/community-builder: add user onny

onny:


zowoq:
@onny You seem to be frequently building risvc packages that take a long time to build. Make sure to use `max-jobs 1` and to not run more than one build at a time.

zowoq:
I've revoked your access for not limiting your builds as asked.

PR #1849, opened 2025-06-01
build03: add freebsd builder

zowoq:


zowoq:
@rhelmot How do you want to set up CI for stdenv, etc? Easiest option might be a release set in nixpkgs, there is already precedent for releases that aren't built by hydra.nixos.org: [pkgs/top-level/release-cuda.nix](https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/release-cuda.nix)

rhelmot:
I'm a bit of a nix ci newbie, so my opinion maybe isn't worth much, but I think a release set in nixpkgs would be awesome. the only thing I think that could be better would be that we could do it under the nixos-bsd namespace for additional flexability, at least for now.

zowoq:
Yes, it would be unlocked, similar to our other nixpkgs jobsets:

https://hydra.nix-community.org/project/nixpkgs
https://github.com/nix-community/infra/blob/master/terraform/hydra-nixpkgs.tf

Occasionally in nixpkgs a mass rebuild is accidentally merged to master instead of staging, usually it'll get reverted quickly but we'd want to follow nixos-unstable-small instead of master to guard against those unnecessary rebuilds being built and cached on our infra.

zowoq:
@rhelmot Your freebsd native stdenv PR is now in staging-next (https://nixpk.gs/pr-tracker.html?pr=389733) but the stdenv doesn't evaluate.

zowoq:
I've deployed the CI builder and pointed the hydra jobset at staging-next, found more breakage unfortunately.

https://hydra.nix-community.org/jobset/nixpkgs/bsd

Will probably need to have a permanent jobset for staging-next as well to catch problems early, I'll need to think a bit as how to set that up to try to limit cache usage.

zowoq:
Moved the hydra and terraform changes to https://github.com/nix-community/infra/pull/1857.

zowoq:
stdenv built on new staging-next successfully: https://hydra.nix-community.org/jobset/nixpkgs/bsd-staging-next.

> Will probably need to have a permanent jobset for staging-next as well to catch problems early, I'll need to think a bit as how to set that up to try to limit cache usage.

Can't really avoid this but shouldn't be much of a problem so won't worry about it.

Would you like to move the release-bsd.nix file to the nixos-bsd org? We could keep it here for now if that is easier?

PR #1868, opened 2025-06-20
{darwin,nixos}/community-builder: add user mdaniels5757

mdaniels5757:
This would be useful to help me test some large builds that exceed GitHub Actions' limits, and to better be able to debug Darwin issues when they arise.

* [x]  Joined https://matrix.to/#/#nix-community:nixos.org


zowoq:
Marking this as a draft as you've only started contributing to nixpkgs last month, we can revisit this in a few of months when you are better known by the community.

PR #1875, opened 2025-06-28
{darwin,nixos}/community-builder: update/add kashw2 user

kashw2:
Darwin:
- Updated my public keys with more machines
- Updated old keys that have been reprovisioned

Nixos:
- Added myself as a user for access to aarch64-linux
- Added my public keys


zowoq:
Currently no one has nushell set as their shell, the last user that tried it reported that it was broken for remote builds: https://github.com/nix-community/infra/pull/1582.

kashw2:
Thanks for the heads up, back to bash it is

PR #1888, opened 2025-07-08
{darwin,nixos}/community-builder: add marcin-serwin

marcin-serwin:
I'm part of the team responsible for maintaining SDL libraries in nixpkgs and sometimes this involves diagnosing issues with  darwin which I don't have easy access to (e.g., https://github.com/NixOS/nixpkgs/issues/407056) and sometimes dealing with cross compilation which takes a long time on my laptop (e.g., https://github.com/NixOS/nixpkgs/pull/418646). Having access to runners would make it easier to deal with such occurrences in the future.


zowoq:
Doing cross builds for a platform that we and the nixos.org support natively isn't a good use of resources, the x86_64-linux builder is the weakest out of all of them.

zowoq:
> Are there some usage guidelines available?

No, this is in the backlog of issues.

> This PR was specifically addressing issues with cross compilation. Doing native aarch64 would not be very interesting since it works already. Should testing cross with the builders be avoided in general or just the x86_64-linux one?

Well, I wouldn't use it for stuff that we already have native builds for. We do have people building cross but that is minimal headless nixos closures, not electron desktop apps.

PR #1891, opened 2025-07-09
modules/nixos/rfc39: add stylix

zowoq:


modules/nixos/rfc39.bash
```diff
@@ -19,6 +19,7 @@ git maintenance run --auto
 repos=(
   "home-manager all-maintainers.nix 13467276"
   "nixvim generated/all-maintainers.nix 13503498"
+  "stylix stylix/generated/all-maintainers.nix 13054517"
```
awwpotato, line 22: we moved it to just `generated/all-maintainers.nix`
```suggestion
  "stylix generated/all-maintainers.nix 13054517"
```

PR #1896, opened 2025-07-16
terraform: generate hydra nixpkgs jobsets with nix

zowoq:
Probably should have just used nix instead of using dynamic blocks in hcl, I'm hoping this is clearer and easier to work with.

I looked at using terranix but wanted to have the json checked in so the full jobset diff is visible.

@Mic92 @zimbatm WDTY?


Mic92:
Not quite sure where the simplification here is, but if you think this is easier? Hydra also has declarative jobsets otherwise where you can use nix directly.

zowoq:
https://github.com/nix-community/infra/pull/1852 is probably a better example, add the jobset in nix with staging_next set and it generates the full json for two jobsets.

> Hydra also has declarative jobsets otherwise where you can use nix directly.

I think keeping it all of it in terraform is easier to work with.

zowoq:
Will stay with just tf files for now.

PR #1955, opened 2025-09-05
modules/nixos/disko-zfs: add dataset for /nix/var/nix/builds

zowoq:
Did this for /tmp in https://github.com/nix-community/infra/pull/1942.

Won't reprovision the machines, I'll just create the dataset manually.

`zfs create -u -o sync=disabled -o mountpoint=legacy zroot/build`

cc @Mic92

### DIFF
--- a/modules/darwin/common/default.nix
+++ b/modules/darwin/common/default.nix
@@ -32,6 +32,17 @@
     HostKey /etc/ssh/ssh_host_ed25519_key
   '';
 
+  launchd.daemons.nix-build-cleanup = {
+    script = "${pkgs.findutils}/bin/find /nix/var/nix/builds -delete || true";
+    serviceConfig = {
+      KeepAlive = false;
+      LaunchOnlyOnce = true;
+      RunAtLoad = true;
+      StandardErrorPath = "/var/log/nix-build-cleanup.log";
+      StandardOutPath = "/var/log/nix-build-cleanup.log";
+    };
+  };
+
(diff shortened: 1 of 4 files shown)


modules/nixos/common/default.nix
```diff
@@ -35,6 +35,8 @@
   zramSwap.enable = true;
   zramSwap.memoryPercent = 100;
 
+  systemd.tmpfiles.rules = [ "D! /nix/var/nix/builds 0755 root root" ];
```
Mic92, line 38: You might want to only do this for old files. Otherwise running a NixOS activation while having a build running produces random crashes.

modules/darwin/common/default.nix
```diff
@@ -32,6 +32,17 @@
     HostKey /etc/ssh/ssh_host_ed25519_key
   '';
 
+  launchd.daemons.nix-build-cleanup = {
+    script = "${pkgs.findutils}/bin/find /nix/var/nix/builds -delete || true";
```
Mic92, line 36: This could take a creation time flag to only run the cleanup for old build dirs.

modules/nixos/common/default.nix
zowoq: This runs on boot, not activation?

modules/darwin/common/default.nix
zowoq: This also runs on boot?

PR #1957, opened 2025-09-06
remove hercules CI

zowoq:


zowoq:
After we've done the x86_64-darwin removal I'll make a post that hercules is going to be removed with 30-60 days notice. IIRC only one repo is using hercules effects for a flake update so I'll replace that with a github app, other projects that are only doing builds should be fine to use buildbot.

zowoq:
> After we've done the x86_64-darwin removal I'll make a post that hercules is going to be removed with 30-60 days notice.

~~Still waiting for x86_64-darwin removal so I'll go ahead with this post in a few days.~~

Going to wait a bit on this, not much point in moving projects from hercules to buildbot and then buildbot to nixbot.

zowoq:
Announced the removal in https://github.com/orgs/nix-community/discussions/2331.

PR #1959, opened 2025-09-06
nixos/community-builder: add sternenseemann

sternenseemann:
Same as https://github.com/nix-community/infra/pull/1028.


sternenseemann:
Actually can't we reuse the key file between the two places?!

zowoq:
> Actually can't we reuse the key file between the two places?!

I've been planning to merge the two lists of users, I've got it half finished on a local branch.

PR #1977, opened 2025-09-19
darwin/community-builder: add adisbladis

adisbladis:


zowoq:
@adisbladis You already have access via customer@darwin01.

zowoq:
Eventually admins will get separate accounts on darwin like we have on linux.

zowoq:
Also see the previous discussion: https://github.com/nix-community/infra/pull/921.

PR #1991, opened 2025-10-04
modules/nixos: move cgroups to common

zowoq:

### DIFF
--- a/modules/nixos/common/cgroups.nix
+++ b/modules/nixos/common/cgroups.nix
@@ -1,6 +1,7 @@
+{ config, lib, ... }:
 {
-  nix = {
-    settings = {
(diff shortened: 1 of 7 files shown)


modules/nixos/common/cgroups.nix
```diff
@@ -1,6 +1,7 @@
+{ config, lib, ... }:
 {
-  nix = {
-    settings = {
+  config = lib.mkIf (lib.hasPrefix "build" config.networking.hostName) {
+    nix.settings = {
       experimental-features = [
         "auto-allocate-uids"
         "cgroups"
```
Mic92, line 7: Did you had any more crashes with this?

zowoq: I'd need to reduce max-silent-time to check for crashes.

PR #2027, opened 2025-11-17
modules/nixos/buildbot: add stylix

danth:
See https://github.com/nix-community/stylix/pull/1961.

In short, there isn't enough disk space on a standard GitHub Actions runner to build all of our checks, especially when starting from an empty cache. This is because each testbed depends on the application it's theming, some of which are quite large. We also can't use a separate runner for each check because there are more checks than the maximum number of parallel jobs.

How much local disk space does BuildBot have? Cache space shouldn't be an issue because most of our dependencies are already on <https://cache.nixos.org>, so wouldn't need to be uploaded to Cachix.

### DIFF
--- a/modules/nixos/buildbot.nix
+++ b/modules/nixos/buildbot.nix
@@ -29,6 +29,7 @@ let
     "nix-community/nixpkgs-xr"
     "nix-community/nixvim"
     "nix-community/srvos"
+    "nix-community/stylix"
     # keep-sorted end
   ];
 


zowoq:
Disk space shouldn't be an issue, the darwin runner has 1TB and linux runners are >= 2TB. We only has one darwin runner to handle both darwin arches so may need to only run tests on aarch64-darwin like home-manger and nixvim.

I've triggered a CI run manually, once it is finished I'll trigger another so we can see a cached CI run.

https://buildbot.nix-community.org/#/builders/6570/builds/1

trueNAHO:
> We only has one darwin runner to handle both darwin arches so may need to only run tests on aarch64-darwin like home-manger and nixvim.
>
> I've triggered a CI run manually, once it is finished I'll trigger another so we can see a cached CI run.

For reference, https://github.com/nix-community/stylix/pull/1968 attempts to fix the legitimately broken CI. In https://github.com/nix-community/stylix/pull/1969, I am experimenting with increasing nothing-but-nix's aggressiveness, otherwise removing the aarch64-linux runner successfully (https://github.com/nix-community/stylix/commit/5602d1a2395895e166f2e9e135441b8aa876b7d4) avoids running out of space and should make our CI pass again.

I will continue investigating https://github.com/nix-community/stylix/pull/1969 tomorrow.

trueNAHO (approved):
With https://github.com/nix-community/stylix/pull/1985 merged, I suppose this here can be merged as well.

0xda157:
@zowoq is there anything we (stylix committers) need to do so buildbot status will be displayed by github?

zowoq:
No. If you rebase a PR or add a commit you'll see it.

PR #2047, opened 2025-12-07
restore build03-freebsd

zowoq:


rhelmot:
Appreciated. Please know that my tenure on this FreeBSD development kick is limited to the current "beach episode" and I will disappear except for being pinged in issues as soon as I'm back to my PhD (ETA: 2 days).

zowoq:
The VMs seem to run okay, evals seem to be much faster now which is a nice improvement.

Builds in the VM aren't working, seems to break trying to build the bootstrap tools.
```
> nix build github:nixos/nixpkgs#stdenv --max-jobs 1

...

error: filesystem error: in copy: Function not implemented ["/etc/resolv.conf"] ["/etc/resolv.conf"]
```

rhelmot:
The writable overlay on the nix store is probably not working correctly. I'll take a look tomorrow.

rhelmot:
What a strange set of commits on your fork. I'll leave as a reference https://github.com/rhelmot/nixpkgs/commit/efa82d388a21c7dc867f118b9891fbf6ff64efa4 and https://github.com/rhelmot/nixpkgs/commit/e175d8ce76d4dd08ac970adf749cb84430c757e5, but I'm not sure they're related.

PR #2070, opened 2025-12-22
docs: switch to zensical

zowoq:
> https://squidfunk.github.io/mkdocs-material/blog/2025/11/11/insiders-now-free-for-everyone/
> As we're shifting our efforts to Zensical, Material for MkDocs is entering maintenance mode. This means that while we'll continue to fix critical bugs and security issues for 12 month at least, no new features will be added to Material for MkDocs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ result-*
 
 # direnv
 .direnv/
+
+# zensical
+site/
--- a/dev/docs.nix
+++ b/dev/docs.nix
@@ -3,15 +3,14 @@
   perSystem =
     { config, pkgs, ... }:
     {
-      devShells.mkdocs = pkgs.mkShellNoCC { inputsFrom = [ config.packages.docs ]; };
+      devShells.zensical = pkgs.mkShellNoCC { packages = [ pkgs.zensical ]; };
       packages = {
         docs =
           pkgs.runCommand "docs"
             {
               buildInputs = [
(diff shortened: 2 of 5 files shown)


mdaniels5757 (changes requested):
I tested this locally with `nix develop .#zensical -c zensical serve"` for every page. Two things I noticed:

* (blocker) The code blocks on the `/infrastructure` page don't render anymore. (They do on the live site.)
* (non-blocker) There is a scroll bar on every page, even when scrolling is not possible because the entire page fits on one screen (e.g. on the homepage). The live site does not render the scroll bar in this situation. Not sure if this is possible or easy to fix, but that'd be nice.

zowoq:
> (blocker) The code blocks on the /infrastructure page don't render anymore. (They do on the live site.)

It's the same with running mkdocs manually on master, the files are generated and copied into the docs build:

https://github.com/nix-community/infra/blob/4e9914095180288eae44839efcaa6ba545c2a508/dev/docs.nix#L27

> (non-blocker) There is a scroll bar on every page, even when scrolling is not possible because the entire page fits on one screen (e.g. on the homepage). The live site does not render the scroll bar in this situation. Not sure if this is possible or easy to fix, but that'd be nice.

I won't link it directly but I think this is issue 92 on https://github.com/zensical/zensical?

mdaniels5757 (approved):
Ah, makes sense. And thanks for digging up that issue: I only looked at the open ones :).

Issue #1926, opened 2025-08-18
increase cache capacity


zowoq:
Currently we have a sponsored 1TB Cachix cache, recently it may only take 1-3 weeks before the LRU paths are evicted.

I think self hosting a cache on a https://www.hetzner.com/dedicated-rootserver/sx65 for ~113€/month is the most cost effective option and it also avoids the bill shock that we may encounter with using hosted storage.

Mic92:
If we combine backblaze s3 with cloudflare, we can get a 6 Dollar
/ TB redundant high-available binary cache. That sound cheaper than paying 113 Euro for a single machine that we cannot reboot without downtime of the cache.

zowoq:
For gha I guess we'll probably want to do something like https://github.com/cachix/cachix-action that sets up the cache, tokens, post-build-hook so users can just add a line to their config and it works OOTB (e.g. `- uses: nix-community/push-to-cache@main`).

wentasah:
Hi, is there any progress? I'm asking because of [nix-ros-overlay migration](https://github.com/orgs/nix-community/discussions/2094).

zowoq:
> Hi, is there any progress? I'm asking because of [nix-ros-overlay migration](https://github.com/orgs/nix-community/discussions/2094).

https://github.com/orgs/nix-community/discussions/2132

I stopped working on this because the future of nix community was unclear.

https://github.com/nix-community/infra/pull/2096#issuecomment-4526988114

I'm resuming work on it, there should be progress by the end of the month.

Issue #532, opened 2023-04-18
removing the org wide cachix secrets


zowoq:
https://github.com/organizations/nix-community/settings/secrets/actions

- CACHIX_AUTH_TOKEN
- CACHIX_SIGNING_KEY

I think we had all agreed these needed to be removed, how to we want to do it, 60 days notice seems reasonable to me?

Projects that want to use the nix-community cachix need to move to hercules or hydra, alternatively they can create their own separate cachix.

This will mean that their is no nix-community cache for darwin as we currently don't have a builder for that platform.

zimbatm:
I deleted the CACHIX_SIGNING_KEY as it's the most critical one and hard to rotate once leaked. I thought I already deleted it. The only impacted project AFAIK is https://github.com/nix-community/hardware-mnt-reform. /cc @jollheef and @ehmry.

For the other token, I propose to make an inventory of the affected repos, and notify the maintainers.

zowoq:
Not as many as I was expecting:

- [dream2nix](https://github.com/nix-community/dream2nix)
- [dream2nix-auto-test](https://github.com/nix-community/dream2nix-auto-test)
- [fenix](https://github.com/nix-community/fenix)
- [home-manager](https://github.com/nix-community/home-manager)
- [lorri](https://github.com/nix-community/lorri)
- [mineflake](https://github.com/nix-community/mineflake)
- [neovim-nightly-overlay](https://github.com/nix-community/neovim-nightly-overlay)
- [nix-init](https://github.com/nix-community/nix-init)
- [nixpkgs-fmt](https://github.com/nix-community/nixpkgs-fmt)
- [nurl](https://github.com/nix-community/nurl)
- [patsh](https://github.com/nix-community/patsh)
- [trustix](https://github.com/nix-community/trustix)

zimbatm:
How reliable and easy to use is Hercules CI? Is it something you would recommend all the repos migrate to?

zowoq:
For standard use (e.g building a flake when a branch is pushed to the repo) I'd say it's fine, I'd expect most repos can migrate without any problems.

Currently it doesn't build PRs from forks so repos would need to use bors, gh merge queue or similar. A repo like home-manager that has a lot of third party contributors may be better off staying on actions with their own cachix so they don't need to deal with that.

Repos that want to properly support darwin (e.g. home-manager) should probably stay on actions anyway, at least until we have a darwin builder.

[...]

PR #2096, opened 2026-01-20
build03: add niks3

zowoq:
https://github.com/nix-community/infra/issues/532, https://github.com/nix-community/infra/issues/1926


zowoq:
I'm planning to resume work in this in the next couple of weeks.

PR #2162, opened 2026-03-12
nixos/community-builder: add miniharinn

MiniHarinn:
I am requesting access to the NixOS Community Builder. This would help me compile large projects like Bambu Studio, GitButler, etc. It would also help with my nixpkgs-review runs.

Thank you for your consideration.


MiniHarinn:
Hi again @zowoq, I've made quite a few contributions to nixpkgs and have lately been fixing many packages in nixpkgs as part of the ZHF campaign. I've found that some of the packages I'm attempting to fix require more computing power/memory than I currently have locally. Could you please reconsider my eligibility for the community builder? (I'm not particularly close to any committers who would go out of their way to vouch for me.)

If this isn't enough to build some trust in me for access to the community builder, I'm happy to keep this as a draft until you find that I'm trustworthy enough :D

MiniHarinn:
Thanks a lot! I'll make sure not to overuse it so other people can still have capacity.

PR #2194, opened 2026-04-12
terrafor/hydra-projects: fetch main branch for nixos-mailserver

mweinelt:
Instead of the default branch hydra defaults to the master branch if no branch is specified. We therefore need to be explicit here.


zowoq:
Applied, looks like it works?

mweinelt:
https://hydra.nix-community.org/jobset/simple-nixos-mailserver/main

Yep, looks good.

PR #2202, opened 2026-04-23
build05: switch to disko-zfs-systemd-boot

zowoq:


mdaniels5757:
Did this wipe my homedir? Looks like I don't have anything older than Apr 23 13:12, which must have been when I re-deployed my personal home-manager setup.

zowoq:
Yes, this was a full reinstall.

PR #2207, opened 2026-04-24
switch to upstream hydra, new queue runner

zowoq:
~~Need x86_64-freebsd support for the hydra builder before we can do this.~~ It builds but haven't tested it yet. Going to wait for h.n.o to deploy the queue runner and for hydra upstream to stabilise a bit as rebuilding the freebsd vm image isn't cheap for the cachix cache.

### DIFF
--- a/dev/packages.nix
+++ b/dev/packages.nix
@@ -9,6 +9,18 @@
     ps.deploykit
     ps.invoke
   ]);
+  nix-eval-jobs = prev.nix-eval-jobs.overrideAttrs (
+    _: p: {
+      version = "2.35.2-unstable-2026-09-01";
+      src = final.fetchFromGitHub {
+        owner = "NixOS";
+        repo = "nix-eval-jobs";
+        rev = "55e658518ae417cf26f36643fcfdebe5c5db17aa";
+        hash = "sha256-4z5GnNd9cbkKChaovYghlxuh1k5rYlxNT7wpZeR1oU0=";
+      };
+      buildInputs = (p.buildInputs or [ ]) ++ [ final.mimalloc ];
+    }
+  );
   nixVersions = prev.nixVersions.extend (
     _: super:
     let
--- a/dev/packages.nix
+++ b/dev/packages.nix
@@ -9,6 +9,26 @@
     ps.deploykit
     ps.invoke
   ]);
+  hydraPackages = {
+    inherit
+      (final.lib.makeScope final.newScope (
+        import "${inputs.hydra}/packaging/components.nix" {
+          version = builtins.substring 0 8 inputs.hydra.lastModifiedDate;
+          releaseVersion = builtins.substring 0 8 inputs.hydra.lastModifiedDate;
+          craneLib = inputs.crane.mkLib final;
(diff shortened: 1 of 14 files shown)


zowoq:
@Mic92 With tribuchet and the new hydra queue runner we'd need to have both deployed on each machine, I was wondering if running a single hydra queue builder with all of our supported systems on the tribuchet hub might work? e.g.

```mermaid
graph TB

    subgraph BUILD03["build03"]
        nixbot["nixbot"]
        hydra["hydra"]
        HQR["hydra-queue-runner"]
        HQB["hydra-queue-builder
        - aarch64-linux
        - aarch64-darwin
        - x86_64-freebsd
        - x86_64-linux
        "]
        TH["tribuchet-hub"]
        build03["x86_64-linux"]
        ND["nix-daemon"]

        nixbot <--> ND
        ND <--> build03
        hydra <--> HQR
        HQR <--> HQB
        HQB <--> ND
        ND <--> TH
    end

    subgraph build04["build04"]
        TW-B04["tribuchet-worker"]
        aarch64-linux["aarch64-linux"]
    end

    subgraph darwin02["darwin02"]
        TW-D02["tribuchet-worker"]
        aarch64-darwin["aarch64-darwin"]
    end

    subgraph build03-freebsd["build03-freebsd"]
        TW-B03F["tribuchet-worker"]
        x86_64-freebsd["x86_64-freebsd"]
    end

    TH <--> TW-B04
    TH <--> TW-D02
    TW-B04 <--> aarch64-linux
    TW-D02 <--> aarch64-darwin
    TH <--> TW-B03F
    TW-B03F <--> x86_64-freebsd
```

Mic92:
When queue builder and runner are collocated on the same machine, they should skip most transfers. I haven't tested that scenario but if it's not already a test case maybe it should be. 

I would maybe wait a little bit longer with deploying tribuchet. It's getting more stable now but I am still discovering bugs through my own CI. I don't have a MacOS instance outside of the github runner test. I planning also on adding a different remote builder client so it doesn't have to displace all builders on a machine and can be more selectively enabled in a CI which would allow to collocate it better with queue builder.

Otherwise it's probably worth thinking of what frontent features are still needed to make nixbot a hydra replacement at least for the non NixOS build infra usecase. The UI scales very well with larger build graphs. It might need an option to show build regressions as a differential few and switch to fully control all derivations scheduled in order to cache failures on all levels - also I am still thinking where the build scheduler should live since tribuchet could also take over this task. I stress tested the build scheduler and it would more likely bottlenecked on external components like tribuchet than it's ability to push out a build graph.

zowoq:
> Otherwise it's probably worth thinking of what frontent features are still needed to make nixbot a hydra replacement at least for the non NixOS build infra usecase.

https://github.com/nix-community/infra/blob/master/terraform/hydra-nixpkgs.tf

What about these hydra legacy type jobsets that follow channels, could that be supported by nixbot or would you only want to support flakes?

Just remembered that the hydra CI limits max jobs, moving those jobsets to nixbot will overload CI unless they have a similar throttle.

PR #2235, opened 2026-05-14
nixos/community-builder: add herbetom

herbetom:
I would like to ask for considerations of granting me access to the community builders.

I'm a maintainer of a couple of packages. I'm currently lacking the ability to run nixosTests on aarch64. In addition i'm currently more or less limited to a 3 core aarch64 build VM which isn't really all that much. This means i'm currently often not doing aarch64 builds since it just isn't feasible. Changing that would be kinda great.


herbetom:
Thanks! :)

PR #2239, opened 2026-05-16
nixos/community-builder: add quantenzitrone

quantenzitrone:
i need to build a nixos test on aarch64-linux to fix the failure
https://hydra.nixos.org/build/328796987


quantenzitrone:
realized the nixos and darwin builders are separated, so no second key needed

PR #2245, opened 2026-05-22
nixos/community-builder: add additional SSH public key for miniharinn + tidy comments

MiniHarinn:
Add one more SSH key for my laptop, and make the comment identifiable in case we need to revoke any of these keys.


MiniHarinn:
Thanks!

PR #2255, opened 2026-05-28
nixos/community-builder: add ysun

stepbrobd:
Hello! Sometimes I need to run VM tests and builds for package bumps on aarch64-linux, binfmt is wayyy to slow

I do have a rpi5 but I'm running a [gps time server](https://time.ysun.co/) on it, put heavy load on it will skew the accuracy quite a bit (well, also its not really fast)

Thus I would like to request access to the aarch64-linux community builder, thanks!


stepbrobd:
Thanks a lot!

PR #2267, opened 2026-06-10
update `nl.meet.nix-community.org` cname

raboof:
bad timing for codeberg pages to be having issues, as we have a meetup later tonight.

set up an alternative, seems OK:

    curl -v --resolve nl.meet.nix-community.org:443:128.140.0.203 https://nl.meet.nix-community.org -k

(except the letsencrypt cert isn't there yet of course)

To make it easier to verify "I'm me": https://codeberg.org/NixNL/ points at https://codeberg.org/raboof which points at engelen.eu in my profile


raboof:
wait something else is going on, CAA-related

PR #2313, opened 2026-07-03
modules/nixos/nixbot: add home-manager

zowoq:
@khaneliman Would you be okay with enabling nixbot alongside buildbot for `home-manager` the same as was done with `nixvim`? Might be easier if we can make sure the repo is running fine on nixbot and moved off buildbot before the repo is moved to the nixos org?

### DIFF
--- a/modules/nixos/nixbot.nix
+++ b/modules/nixos/nixbot.nix
@@ -7,12 +7,14 @@
 let
   repoAllowlist = [
     # keep-sorted start case=no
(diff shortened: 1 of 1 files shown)


khaneliman:
Sounds good to me

zowoq:
@khaneliman Any issues with nixbot? Are you okay with dropping buildbot? https://github.com/nix-community/infra/pull/2314

PR #2317, opened 2026-07-03
darwin/community-builder: add hythera

Hythera:
With `x86_64-darwin` going out of support now I don't have a reliable way to test and debug package updates / fixes on `darwin` anymore. It has mostly come very much in handy when doing packgage updates that cause mass-rebuilds.
Access to the darwin community builder would allow me to build, test, debug package updates of PRs I review / my own ones again.


Hythera:
Thanks. Are there any up-to-date docs on how many  `cores` and `max-jobs` one is supposed to use, especially when doing `stdenv` rebuilds?

zowoq:
No. This machine is already overloaded so max-jobs 1 and cores 2. I wouldn't use it for stdenv rebuilds.

PR #2335, opened 2026-07-11
docs: remove links to deleted repo

phanirithvij:
@ckiee if your repo is still public somewhere or you could make these resources available via a public or secret gist, please open a PR with the correct resources.


ckiee:
Hi @phanirithvij, for sure: https://gist.github.com/ckiee/f7768f55a938aa7c985406f209a6f114

Turns out that beyond just the repo being private now, this commit was also rebased out of existence :P

PR #2338, opened 2026-07-12
docs: add example home-manager url

phanirithvij:
recently removed in #2335

Link is from the comment https://github.com/nix-community/infra/pull/2335#issuecomment-4944205729


zowoq:
Thanks, I'll merge this as-is but might be nice to have a simpler example.

PR #2358, opened 2026-07-25
move a bunch of repos from {buildbot,hercules} to nixbot

figsoda:
i can disable the hercules repos once this is merged


zowoq:
Thank you!

PR #2376, opened 2026-08-01
Revert "use ssh knownHosts for host sops keys"

zowoq:
Switching back to age keys so we can follow upstream instead of a fork.


trueNAHO:
FYI, https://github.com/FiloSottile/age recently added support for post-quantum encryption.

zowoq:
> FYI, https://github.com/FiloSottile/age recently added support for post-quantum encryption.

We're probably due to rotate our secrets, I'll see about doing this at the same time.

PR #2399, opened 2026-08-15
{darwin,nixos}/community-builder: add user thunze

thunze:
- `aarch64-linux` access would be useful for testing builds that GHA cannot handle, e.g. due to size or features that break on GHA (e.g. bwrap). It would also be useful for debugging aarch64-specific issues that surface from time to time.
- `aarch64-darwin` access would be great for debugging darwin-specific issues, which I can no longer reliably do now that `x86_64-darwin` support has been discontinued.

Thanks!


modules/darwin/community-builder/keys/thunze
```diff
@@ -0,0 +1,3 @@
+sk-ecdsa-sha2-nistp256@openssh.com AAAAInNrLWVjZHNhLXNoYTItbmlzdHAyNTZAb3BlbnNzaC5jb20AAAAIbmlzdHAyNTYAAABBBGV7njWQY/hN441NOwz/jdUbLQElxv4Wz6qHPeA7a7uZ7nOwVniX7hTYpdg7A8E4gLxKSG6+ASQZvjJ6bXcxM1cAAAAEc3NoOg==
+sk-ecdsa-sha2-nistp256@openssh.com AAAAInNrLWVjZHNhLXNoYTItbmlzdHAyNTZAb3BlbnNzaC5jb20AAAAIbmlzdHAyNTYAAABBBOmj013XhPlPdsc89hieBI0pr5ZZTKvxd7CuYPnyOq+jdohL0koG6hFkdAQJ7Ild6FWHdgmeW4zAQM9/pZyy5mcAAAAEc3NoOg==
+ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIIxT3kard/+nPPZAfav2wPFUt0hJDoAA+LSv2dHVJIy3
```
thunze, line 3: The darwin builder still doesn't support `sk-*` keys, correct? Otherwise I'd remove the extra ed25519 key here. If the builder doesn't support `sk-*` keys, and if it's OK, I'd still add them in case `sk-*` keys become supported in the future (except if that definitely won't happen).

PR #2439, opened 2026-09-01
Add viraptor to darwin users

viraptor:
I'd like to be able to build a few packages away from darwin hardware I have access to. I'm working on some changes which end up redoing darwin bootstrap, but a few packages (especially the linker) keep triggering my enforced security suite. I'd like to wrap up the work on https://github.com/NixOS/nixpkgs/issues/512622 now and offloading those problematic packages to the community builder would really help me out.


viraptor:
That was quick, thank you ❤️

